PR #59, opened 2019-07-01
Add hammerdbrun to run command line scripts

xychar:
Add command line tool to run command line scripts directly instead of typing commands by hand.
This is mainly for Windows, as using pipe for standard input does not work on Windows.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -860,3 +860,9 @@ Bug fixes
 [#131] Duplicate Files with incorrect prefix in MySQL and Redis Directories
 
 Added Oracle Database Metrics based on Ashmon by Kyle Hailey, used with permission to add as GPLv3
+
+---------------------------------------------------------------------
+
+Version 3.1a Jul 2019
+
+Add command line tool hammerdbrun to run command line scripts
--- a/hammerdbrun
+++ b/hammerdbrun
@@ -1,0 +1,67 @@
+#!/bin/sh
+#########################################################################
+## \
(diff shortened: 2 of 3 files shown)


xychar:
Thank you @sm-shaw, you are right, I just copied the thought from the GUI app, I often  run the scripts remotely via SSH, the console version is slim and good enough. And also I can combine tasks in command line parameters, e.g. sometimes I only need create schema, but I may want run the workload following it.

xychar:
It would be better to add the functionality to hammerdbcli, not necessary to add another command line entry, I guess.

xychar:
If there are command line parameters, run the scripts and exit, otherwise go to interactive mode, that would be perfect.

sm-shaw:
OK v3.2 has been now been pushed to master - hammerdbcli and hammerdbcli.bat have been changed to add the functionality for this pull request in a different way . (hammerdb.bat has been changed as well to pass the arguments through on windows). With these versions what you need to do is something similar to the GUI such as:

`hammerdbcli auto script_to_run.tcl`

Please note that there is one difference between running a script in interactive mode and running a script directly like this. In interactive mode it has been set to enter the event loop automatically and therefore any script passed at the command line through a pipe will do this for you.  (The GUI also automatically enters the event loop). Passing a script this way will not enter the event loop automatically and therefore what you will need to do is add the following line to the end of the script you are going to run to do this.

`vwait forever`

Otherwise (because its multi-threaded) by default once it reaches the end of the script then it will exit.

xychar:
That's great, thank you Steve.

PR #60, opened 2019-07-11
Fix order of arguments to do_tpcc in pgoltp.tcl

marcocitus:
The $pg_dbase argument is in the wrong place causing errors when running against PostgreSQL (the password is used as the username).


sm-shaw:
I did some testing to verify and yes this  was the case and the commit fixes the issue so the pr has been merged - many thanks

PR #69, opened 2019-07-20
Allow passing in extra parameters for the user script

xychar:
I changed the command line parameter validation, to allow passing in extra parameters, so in the user script, we can get more flexibility to make use of the extra parameters, instead of changing user script every time.

For example, I can use it like this:
`bin\tclsh86t.exe auto my-script.tcl server=server1 port=5432`

### DIFF
--- a/hammerdb
+++ b/hammerdb
@@ -35,7 +35,7 @@ set UserDefaultDir [ file dirname [ info script ] ]
 namespace eval autostart {
     set autostartap "false"
     if {$argc == 0} { ; } else {
-    if {$argc != 2 || [lindex $argv 0] != "auto" } {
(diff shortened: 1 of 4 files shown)


xychar:
Also, I suggest to change the .bat file, do not use "START" and "exit" in the batch file.
I would be a bit annoying when I run it from command line, and it will close my current console.

Sometimes, I also want to run it in the same console window instead opening a new one.

xychar:
Yes, the colors also cause troubles when parsing the output by other tools. Anyway, will run the exe directly.

sm-shaw:
OK - lets not go ahead with this for the time being. Both are certainly reasonable suggestions however the default behaviour currently supported in both cases is the most failsafe way.

PR #91, opened 2020-01-14
Added Master/Slave Modes to CLI

sm-shaw:
These updates add the the Master Slave Modes feature to the CLI (https://www.hammerdb.com/docs/ch12.html.)  Setting an instance of HammerDB as a Master enables connecting other instances as Slaves that are then driven by the commands from the Master. Feature is typically used for testing virtualization and clustering.

### DIFF
--- a/hammerdbcli
+++ b/hammerdbcli
@@ -51,7 +51,7 @@ exit
       }
 }
 
-append modulelist { Thread msgcat xml tclreadline }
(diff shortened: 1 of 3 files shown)


abondvt89:
This pull request was accepted as a passed motion of the TPC OSS subcommittee on January 21, 2020.

PR #92, opened 2020-01-18
Refactor Postgres procedural code for performance

daveweb39:
Rewrite of the functions and stored procedures using bulk processing and more RETURNING functionality.  The approach taken is based on that the in Oracle TPC-C Full Disclosure report: http://c970058.r58.cf2.rackcdn.com/fdr/tpcc/Cisco-Oracle%20C240%20TPC-C%20FDR.pdf

On a CPU bound system, testing reveals increase in NOPM:
PG10: 51753 to 69200 (34%)
PG11: 62131 to 74166 (19%)

Whilst every effort has been made to verify no regression in performance, the lack of full regression pack makes it hard to be certain but the workload appears to be equivalent.

Note that one issue was "fixed" i.e. non local order distribution.  I don't believe this has any impact of workload but does adhere to the TPC-C specification.

### DIFF
--- a/src/generic/gentpcc.tcl
+++ b/src/generic/gentpcc.tcl
@@ -15,7 +15,7 @@ set wh [expr ((($myposition - 2) + ($addMore * $loadUserCount))%$w_id_input)+1]
 if {$addMore >= $whRequiredCount} {
 set addMore -1
 } else {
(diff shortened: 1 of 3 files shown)


abondvt89:
Pull Request unanimously approved by vote of the TPC-OSS subcommittee on February 4th, 2020

PR #97, opened 2020-02-07
Partitioning for Oracle ORDERS and HISTORY

sm-shaw:
Testing shows between 1-2% improvement at high performance levels by reducing buffer busy waits


abondvt89:
Accepting pull request as voted on February 18, 2020 in the TPC-OSS subcommittee.

Issue #99, opened 2020-02-20
Reverse print out of metrics to put NOPM ahead of TPM  for version compatibility


sm-shaw:
Currently HammerDB prints out a test result as follows: 
`Vuser 1:TEST RESULT : System achieved 3370315 PostgreSQL TPM at 1465149 NOPM`
where TPM is the user commits + user rollbacks from the database system metrics and NOPM the new orders per minute taken from the schema. 
suggested minor and simple change to reverse this print out to: 
`System achieved 1465149 NOPM at 3370315 PostgreSQL TPM`
The reason is to ensure ongoing compatibility given already committed (and potential future) changes to the stored procedures that changes the ratio between TPM and NOPM to ensure that the user is aware that NOPM gives the most consistent metric between releases and TPM should not be used to compare between versions.  For example between v3.2 and v3.3 when updating the SQL Server driver interface, script and correcting the NUMA imbalance and additional separate commit was removed from the driver script after each stored procedure call - this reduced the TPM however the NOPM (and the workload) stays the same.  For commited changes to the Oracle stored procedures at the next release testing shows the TPM remains similar but bulk operations increases the NOPM. Also if the user is running additional work on the system unrelated to HammerDB that will also be captured in the TPM metrics. TPM remains useful as it relates best to gathered database statistics and can be used in the graphical transaction counter without impacting the test in progress. However NOPM is the best metric for the performance of the HammerDB workload itself. Reversing the print out puts the most important metric first when updating the drivers.

PR #104, opened 2020-02-29
Reverse print out of metrics to put NOPM ahead of TPM

sm-shaw:
Detailed in Issue #99 - reports test results as NOPM from TPM


abondvt89:
Accepting pull request as voted on March 3, 2020 in the TPC-OSS subcommittee.

Issue #100, opened 2020-02-20
Move redis workload to unsupported status


sm-shaw:
The redis implementation in HammerDB stands apart from the other databases in being non-relational  with the application logic in the client.  Also being a single threaded database Redis does not scale for OLTP workloads and is not considered fair to compare to a full database: ie https://redis.io/topics/benchmarks 
Redis is, mostly, a single-threaded server from the POV of commands execution (actually modern versions of Redis use threads for different things). It is not designed to benefit from multiple CPU cores. People are supposed to launch several Redis instances to scale out on several cores if needed. **It is not really fair to compare one single Redis instance to a multi-threaded data store.**
Redis also does not support analytic workloads. 
Suggest to move Redis to unsupported at next release so it does not show as an option in the GUI by default but will remain to be enabled if required. This enables better focus on the currently supported top 5 relational databases in the db-engines ranking https://db-engines.com/en/ranking

This change simply requries moving the comment line in database.xml above the redis configuration with consideration for removal to be decided at a future point in time.

`<!--Trafodion deprecated, uncomment to enable as unsupported`

PR #105, opened 2020-02-29
Deprecated Redis

sm-shaw:
Detailed in issue #100 commented out Redis so it doesn't show in the default database list due to single threaded performance bottleneck. User can still uncomment to use unsupported. Will not develop further features such as graphical statistics for unsupported workloads.


abondvt89:
Accepting pull request as voted on March 3, 2020 in the TPC-OSS subcommittee.

PR #98, opened 2020-02-18
39 MySQL prepared statements

sm-shaw:
Adds user configurable option to use server side prepared statements in the driver script.  Also adds option to connect by socket when testing on a local server.

PR #95, opened 2020-01-22
PostgreSQL prepared statements for functions and tablespace option for schema builds

sm-shaw:
Add prepared statement functionality for PostgreSQL as per Issue #94

PR #114, opened 2020-03-30
Redo PostgreSQL tablespace changes from #95 in pgopt.tcl

marcocitus:
#98 accidentally reverted the tablespace changes to `pgopt.tcl` made by #95, which causes TPC-C on PostgreSQL to fail with `Error: can't read "pg_tspace": no such variable`

This PR redoes the changes from #95 to fix the issue. There were some [minor changes to pgopt.tcl](https://github.com/TPC-Council/HammerDB/pull/98/files#diff-ddb170835ad5c6cdafd7722c4e0623b1R340) in #98 related to handling of `pg_async_scale`. I left those unchanged.

### DIFF
--- a/src/postgresql/pgopt.tcl
+++ b/src/postgresql/pgopt.tcl
@@ -151,7 +151,7 @@ upvar #0 configpostgresql configpostgresql
 setlocaltpccvars $configpostgresql
 #set matching fields in dialog to temporary dict
 variable pgfields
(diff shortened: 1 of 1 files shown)


sm-shaw:
Many thanks for picking this up - yes you are absolutely correct.  As background working on #98 I noticed a minor issue in the MySQL GUI options that asynchronous scaling options were not greyed out when the options window is first shown and test showed that this was the same for a number of the databases. (ie in v3.3 when you click on timed driver script it enables all of the asynchronous options which it shouldn't do). As a quick fix I made what I thought was a small change included with the pull request to the other database options. However as you point out on this branch I was working on I must have omitted to bring it up to date thereby overwriting the GUI options in postgresql for the tablespace when the pull request was committed.  As you point out your changes now restore these and also leave the correction of the greying out of the options in place.  Thanks again for highlighting this and best practice noticed not to add quick fixes to other pull requests!

Issue #118, opened 2020-04-06
HammerDB CLI issue with Changing MsSQLs ODBC Driver


ivedw:
When trying to replicate the GUI workflow I have used on the CLI, when I try to change the driver in the connection setting, with the command:

diset connection mssqls_odbc_driver {ODBC Driver 17 for SQL Server}

I get the following error:

Clearing Script, reload script to activate new setting
Script cleared
Error: Driver script must be either "test" or "timed"

I noticed this is same error given when I try to change the value in the "tpcc" part of the dict that is similar: "mssqls_driver". 

I think the problem might be that the variable checking logic from the tpcc driver item might be crossed with the connection part of the dict.

Thanks

sm-shaw:
Hi, yes confirmed as a bug.  It is checking if the driver has been changed by matching "*driver" and then picking this up for this particular variable that clashes.  A fix will be added to the next version. As a workaround you can run the following instead and set the value directly  (it is advisable to drag the command window out a big larger so the line does not overlap when typing). 
```
hammerdb>dict set configmssqlserver connection mssqls_odbc_driver {ODBC Driver 17 for SQL Server}
[...]

PR #119, opened 2020-04-07
Fix for Issue #118 using diset with variables with spaces

sm-shaw:
Fix for Issue #118. Previous version mismatched odbc_driver as driver script variable and also omitted multiple values separated by spaces. This version fixes both issues in the CLI and tested with new feature of Master/Slave in the CLI to ensure values with spaces are correctly translated to the remote instance of HammerDB.


abondvt89:
Accepting pull request as voted on April 7, 2020 in the TPC-OSS subcommittee.

PR #122, opened 2020-05-13
Add pg_createuserdatabase option to skip CreateUserDatabase if false

marcocitus:
This PR adds a `pg_createuserdatabase` setting to the PostgreSQL benchmarks, which controls whether HammerDB should create a user and database and defaults to `true` to preserve the current behaviour by default.

The current behaviour of always creating a user and a database is inconvenient in several scenarios.

For instance, most managed PostgreSQL services do not provide superuser access to their customers because the superuser can run shell commands on the host. That also means that creating users and/or databases directly is not allowed in some managed services.

If the user and/or database already exist, then the only way to make the build step in HammerDB work is to drop both of them. Hence there is no way to preserve other tables, settings, objects, or PostgreSQL extensions across rebuilds. (Note that databases in PostgreSQL are completely isolated, so you cannot do queries across databases)

We usually comment out the call to `CreateUserDatabase` before starting HammerDB, so it would be quite convenient if it was a setting that we could disable.

### DIFF
--- a/config/postgresql.xml
+++ b/config/postgresql.xml
@@ -19,6 +19,7 @@
             <pg_dritasnap>false</pg_dritasnap>
             <pg_oracompat>false</pg_oracompat>
             <pg_storedprocs>false</pg_storedprocs>
+            <pg_createuserdatabase>true</pg_createuserdatabase>
         </schema>
         <driver>
             <pg_total_iterations>1000000</pg_total_iterations>
@@ -48,6 +49,7 @@
 	    <pg_tpch_gpcompat>false</pg_tpch_gpcompat>
 	    <pg_tpch_gpcompress>false</pg_tpch_gpcompress>
             <pg_num_tpch_threads>1</pg_num_tpch_threads>
+            <pg_createuserdatabase>true</pg_createuserdatabase>
         </schema>
         <driver>
             <pg_total_querysets>1</pg_total_querysets>
--- a/src/postgresql/pgolap.tcl
+++ b/src/postgresql/pgolap.tcl
@@ -521,7 +521,7 @@ error "[pg_result $result -error]"
 return
 }
 
-proc do_tpch { host port scale_fact superuser superuser_password defaultdb db tspace user password greenplum gpcompress num_vu } {
+proc do_tpch { host port scale_fact superuser superuser_password defaultdb db tspace user password createuserdatabase greenplum gpcompress num_vu } {
 global dist_names dist_weights weights dists weights
 ###############################################
 #Generating following rows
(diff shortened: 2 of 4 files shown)


sm-shaw:
I have reviewed the code, the feature and the changes look fine. I will test it prior to the next TPC-OSS meeting on 19th May so it can be voted on for acceptance and merged next week.

sm-shaw:
The intention of the changes are fine in principle however this causes an error when loading the files running the GUI with the file pgoltp.tcl. The error is an additional close curly bracket so the file does not parse correctly and prevents the PostgreSQL schema build as the build_pgtpcc is  not  created as a command. 

There are a couple of options to approach this better. Firstly you should put all of the changes inside the CreateUserDatabase proc and pass the additional option into here. At its simplest you you check if this option is set to skip the build it just prints a message notifying that the user and database have not been created and then returns, by default it would run the user and database creation as normal.  Also doing this in pgopt.tcl it would be best to reverse the logic of "create user/database" to something like "skip create user/database" so that the checkbox by default is unchecked and the user and database are created without the person running it changing the options. 
Another option and you can verify if this works for what you intend in the SQL Server build it does something similar automatically however in this case there is no additional option, instead it runs a SQL query to check if the database already exists under the user given and is empty. In this case it does the schema build against the empty pre-created database. If tables already exists under the user it errors - this way you can use an empty precreated user and database. If this is what is intended then this way is better as there is no changes to the options and again you only have to change the CreateUserDatabase proc by adding the SQL check to run as the tpcc user and password given. 

Also the option pg_createuserdatabase was added to the test driver script. This is not needed in the driver script only in the build script which you modified previously. However for reference any valid changes to the driver script need to be added to the test and timed/timed asynchronous driver scripts as well.

marcocitus:
> The error is an additional close curly bracket so the file does not parse correctly and prevents the PostgreSQL schema build as the build_pgtpcc is not created as a command.

Sorry, I had fixed this locally, but had neglected to push it. Pushed it now.

> Another option and you can verify if this works for what you intend in the SQL Server build it does something similar automatically however in this case there is no additional option, instead it runs a SQL query to check if the database already exists under the user given and is empty. In this case it does the schema build against the empty pre-created database. If tables already exists under the user it errors - this way you can use an empty precreated user and database. If this is what is intended then this way is better as there is no changes to the options and again you only have to change the CreateUserDatabase proc by adding the SQL check to run as the tpcc user and password given.

I think that makes sense. Will update the PR to that effect.

sm-shaw:
Many thanks that sounds like a good plan  - the example from SQL Server is below so firstly it checks if the database is there and if it is then it checks if it is empty to prevent accidentally running 2 builds on the same schema. 
```
proc CreateDatabase { odbc db imdb azure } {
set table_count 0
puts "CHECKING IF DATABASE $db EXISTS"
set rows [ $odbc allrows "IF DB_ID('$db') is not null SELECT 1 AS res ELSE SELECT 0 AS res" ]
set db_exists [ lindex {*}$rows 1 ]
if { $db_exists } {
if {!$azure} {$odbc evaldirect "use $db"}
set rows [ $odbc allrows "select COUNT(*) from sys.tables" ]
set table_count [ lindex {*}$rows 1 ]
if { $table_count == 0 } {
puts "Empty database $db exists"
```

marcocitus:
Will open a new PR once I've made the changes.

Issue #123, opened 2020-05-15
Add a CH-benCHmark (HTAP) workload


marcocitus:
We often run HammerDB concurrently with the analytical queries from [CH-benCHmark](https://db.in.tum.de/research/projects/CHbenCHmark/) (mainly against PostgreSQL) by using a separate script.

Would there be interest in adding a CH workload directly to HammerDB? This involves running TPC-C transactions and TPC-H-like analytical queries concurrently on the same tables. We may be able to write and contribute a PostgreSQL implementation.

CH-benCHmark is also supported by some other benchmarking tools, e.g.: https://github.com/oltpbenchmark/oltpbench

abondvt89:
This is an interesting idea regarding adding it to HammerDB as a combo workload option.  If you have an implementation that you would like to contribute for PostgreSQL that is not license restricted and meets the TPC Fair use guidelines (http://www.tpc.org/tpc_documents_current_versions/pdf/tpc_fair_use_quick_reference_v1.0.0.pdf) go ahead and submit a PR.

sm-shaw:
As discussed this is a great suggestion for inclusion in HammerDB. Before submitting a PR for PostgreSQL it would be good to discuss the design approach to make sure that it is as straightforward to implement as possible for multiple databases.  If the CH workload is sufficiently different from both C and H it would be best to add it as a separate workload. This would involve adding additional parameters to the XML, options to the opt files eg pgopt.tcl and another workload releated script for build and driver such as pghtap.tcl.  To do so would mean more generic work underneath but could pay off in the longer run. If it is very close to either C or H then it could be added as a checkbox option. With a checkbox option the approach is to try and keep the driver scripts as consistent as possible over time, therefore when we have changes such as 'use all warehouses' or 'time profile' they are added by inline editing of the driver script. (There are 3 templates test, timed and asynchronous timed) As these checkbox options increase it is important to ensure that they are all compatible in different combinations.  The multiple labelled connections https://github.com/TPC-Council/HammerDB/issues/101 also takes this editing approach.  Therefore the first task is to determine how sufficiently different this workload is from C & H and whether a separate workload is best not to confuse people running HammerDB for the first time.

sm-shaw:
Adding link to discussion topic [Vector Search Benchmarking along with OLTP ](https://github.com/TPC-Council/HammerDB/discussions/716) There is potential for this to be developed a TPC-CHV workload for databases that support it.

sm-shaw:
[...]

PR #124, opened 2020-05-15
Swap column order in stock_i1 index in pgoltp.tcl

marcocitus:
This PR changes the column order in the stock_i1 index in pgoltp.tcl to be consistent with [oraoltp.tcl](https://github.com/TPC-Council/HammerDB/blob/master/src/oracle/oraoltp.tcl#L1050).

This change does not affect TPC-C performance, but we often run HammerDB concurrently with the analytical queries from [CH-benCHmark](https://db.in.tum.de/research/projects/CHbenCHmark/) and see poor query plans on some PostgreSQL systems causing a single query to dominate the runtime. One the PostgreSQL committers on our team recommended the order of the stock_i1 index to improve the plans, which can make Q2 up to 500x faster without slowing down other queries or affecting TPC-C performance.

Related to #123


sm-shaw:
Yes also the changes look fine to me. I will test it along with the other pull request prior to the next TPC-OSS meeting on 19th May so it can be voted on for acceptance and merged next week.

sm-shaw:
I have run tests both with the default column ordering and with the index column order reversed. The peak result was within 0.1% so it does not seem to impact performance but re-checking the higher results where there was some variance but should not be expected as due to the change. 

columorder2

Also noting that this column order for Stock is used for SQL Server, Oracle and Db2 

SQL Server - CONSTRAINT [stock_i1]  PRIMARY KEY NONCLUSTERED HASH ( [s_i_id], [s_w_id])
Oracle - CREATE UNIQUE INDEX STOCK_I1 ON STOCK (S_I_ID, S_W_ID) INITRANS 4 PCTFREE 1
Db2 - ORGANIZE BY KEY SEQUENCE ( S_I_ID STARTING FROM 1 ENDING AT 100000, S_W_ID STARTING FROM 1 ENDING at $count_ware )

but not for MySQL, or PostgreSQL

MySQL - PRIMARY KEY (`s_w_id`,`s_i_id`)
PostgreSQL - ALTER TABLE STOCK ADD CONSTRAINT STOCK_I1 PRIMARY KEY (S_W_ID, S_I_ID)

sm-shaw:
I rechecked by rerunning with the changed column order and this is all good in terms of performance

columnorder3

sm-shaw:
Accepting pull request as voted on May 19, 2020 in the TPC-OSS subcommittee.

PR #132, opened 2020-06-25
Use Existing User and DB in Postgres

halilozanakgul:
Adds support for using existing user and database in Postgres.

Instead of the current behavior of always creating user and database, this checks if the user or database exists and uses them if they do. The database is only used when there are no tables in it. If there are tables, an error is given.

This PR is opened for the changes discussed in #122.

### DIFF
--- a/src/postgresql/pgolap.tcl
+++ b/src/postgresql/pgolap.tcl
@@ -68,14 +68,42 @@ pg_result $result -clear
 return $lda
 }
 
-proc CreateUserDatabase { lda db tspace superuser user password } {
-set stmnt_count 3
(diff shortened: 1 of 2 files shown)


sm-shaw:
From reviewing the code this looks fine, the approach is good. I will test it for good measure. The TPC-OSS will vote on it and then the plan is to include it in the upcoming release.

sm-shaw:
I have tested this for both the tpc-c and tpc-h schemas and it all works fine so recommending for inclusion in the upcoming release.

sm-shaw:
Accepting pull request as voted on Jul 07, 2020 in the TPC-OSS subcommittee.

Issue #134, opened 2020-06-27
PostgreSQL's neword() function often causes infinite loop


koreshiki:
(I'm using machine translation and my English may be wrong, sorry.)

Hello,
I wonder PostgreSQL's neword() function causes infinite loop under below condition.

Conditions:
1. no_w_id = 1
2. no_max_w_id = 2
3. x = 1

where, no_w_id and no_max_w_id are first and second argument of neword() respectively, and x is variable declared in neword().

Problem code is below.
```
x := round(DBMS_RANDOM(1,100));
IF ( x > 1 )
THEN
no_ol_supply_w_id := no_w_id;
ELSE
no_ol_supply_w_id := no_w_id;
--no_all_local is actually used before this point so following not beneficial
no_o_all_local := 0;
WHILE ((no_ol_supply_w_id = no_w_id) AND (no_max_w_id != 1))
LOOP
no_ol_supply_w_id := round(DBMS_RANDOM(1,no_max_w_id));
END LOOP;
```

Loop operation will be executed if ELSE clause executed (meaning x = 1) and no_max_w_id != 1.
Because function RDBMS_RANDOM(a, b) (assume a > b) returns integer N (a <= N < b), RDBMS_RANDOM(1, 2) returns only 1.
Threfore, this operation results in infinite loop because no_ol_supply_w_id is always 1.

How to repeat:

I used
Windows10 Home(Host)
CentOS8.1 (Guest)
Oracle Virtualbox 6.1
HammerDB3.3(Linux-64bit tar file)
PostgreSQL12.3

```
hammerdb>dbset db pg
hammerdb>dbset bm tpc-c

hammerdb>diset tpcc pg_count_ware 2
hammerdb>diset tpcc pg_num_vu 2
hammerdb>diset tpcc pg_driver timed
hammerdb>diset tpcc pg_timeprofile true

hammerdb>buildschema
hammerdb>vudestroy

hammerdb>loadscript

hammerdb>vuset vu 32
hammerdb>vuset logtotemp 1
hammerdb>vuset unique 1
hammerdb>vuset timestamps 1

hammerdb>vucreate
hammerdb>vurun
```

Then you will get below results.(lock_timeout = 0, statement_timeout = 0)

```
# SELECT
  pid, 
  query, 
  now() - query_start AS duration 
FROM
  pg_stat_activity
WHERE
  state = 'active' 
AND
  pid <> pg_backend_pid()
ORDER BY 
  duration DESC 
[...]

PR #136, opened 2020-06-30
update dbms_random for PostgreSQL neword infinite loop

sm-shaw:
Fix for bug described in Issue #134


abondvt89:

PR #139, opened 2020-07-07
Fix XML closing tags for sprocs and connpool

sm-shaw:


abondvt89:

PR #158, opened 2020-10-22
Fixed statistics updates failure on AWS

raymond13513:
For sql server on aws-rds, sp_updatestats fails due to permission error. Here is an article discussing this issue. https://yatb.mitza.net/2017/03/running-spupdatestats-on-aws-rds.html

The error message looks like this when I run tpch or tpcc on aws-rds
```
Vuser 1:UPDATING SCHEMA STATISTICS

Error in Virtual User 1: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]User does not have permission to perform this action.
(executing the statement)

Vuser 1:FINISHED FAILED

ALL VIRTUAL USERS COMPLETE
```

The work around is to create a stored procedure which wraps sp_updatestats and run it under a different user.


sm-shaw:
Accepting pull request as voted on Oct 27, 2020 in the TPC-OSS subcommittee.

PR #166, opened 2020-11-13
Add is azure flag so hammerdb works on Azure

raymond13513:
This is contains the change need for making Hammerdb works on Azure.
The major changes are:

Include pg_azure flag, similar to mssql
Update ConnectToPostgres to include special case where the azure flag is set to true
a) Make sure the user contains @machine_name
b) Enable SSL connection if is_azure is set to True
Update CreateUserDatabase to handle @machine_name, since this is not allowed in postgres syntax, it will throw error. The @machine_name suffix is not needed after logging in as well. So remove the @

I verified that both TPCC and TPCH works on Azure after setting pg_azure to true.


sm-shaw:
Many thanks for the updated pull request. I will test this against a standard build and have requested input for testing against Azure.

src/postgresql/pgolap.tcl
```diff
@@ -50,16 +50,31 @@ pg_result $result -clear
...
+    if { $azure eq "true" } {
+        set contains_machine_name [string match "*@*" $user]
+        if {$contains_machine_name == 0} {
+            set machine [lindex [split  "$host"  .] 0]
+            set user "$user@$machine"
+        }
+    }
+
+    #Azure requires ssl connection
+    set sslConnectionEnabled 0
```
marcocitus, line 64: Do we want to always disable SSL?

src/postgresql/pgolap.tcl
```diff
@@ -50,16 +50,31 @@ pg_result $result -clear
...
+    if { $azure eq "true" } {
+        set sslConnectionEnabled 0
```
marcocitus, line 66: This sets sslConnectionEnabled to 0 again, which seems like a noop.

In general, it seems to connect without the ssl-related changes.

src/postgresql/pgolap.tcl
```diff
@@ -68,14 +83,21 @@ pg_result $result -clear
...
 set stmnt_count 1
 puts "CREATING DATABASE $db under OWNER $user"
 set result [ pg_exec $lda "SELECT 1 FROM pg_roles WHERE rolname = '$user'"]
 if { [pg_result $result -numTuples] == 0 } {
 set sql($stmnt_count) "CREATE USER $user PASSWORD '$password'"
 incr stmnt_count;
+        if { $azure eq "true" } {
+            # After logging in to postgres
+            # we do not need the machine name
+            set superuser [lindex [split  "$superuser"  @] 0]
```
marcocitus, line 96: It's slightly odd that we first append the host, and then later remove it here. A separate argument might be cleaner. An alternative would be to use "user@host" as the configuration value, and only strip the hostname in this function.

src/postgresql/pgolap.tcl
```diff
@@ -738,6 +760,7 @@ set degree_of_parallel \"$pg_degree_of_parallel\" ;# Degree of Parallelism
 set scale_factor $pg_scale_fact ;#Scale factor of the tpc-h schema
 set host \"$pg_host\" ;# Address of the server hosting PostgreSQL
 set port \"$pg_port\" ;# Port of the PostgreSQL Server
+set azure \"$pg_azure\";#Azure Type Connection
```
marcocitus, line 763: if we end up adding this option, should it be in the UI as well?

marcocitus (commented):
When trying to run this against a single server on Azure, I still get an error:
```
Error in Virtual User 1: ERROR:  must be superuser to alter replication users
```

src/postgresql/pgoltp.tcl
```diff
@@ -1595,14 +1610,21 @@ pg_result $result -clear
 return $lda
 }
 
-proc CreateUserDatabase { lda host port db tspace superuser superuser_password user password } {
+proc CreateUserDatabase { lda host port azure db tspace superuser superuser_password user password } {
```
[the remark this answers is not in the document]
sm-shaw: There is already #132 by @halilozanakgul that gave the functionality suggested in #122. This doesn't skip CreateuserDatabase completely but checks if the databases exists and is empty and if it is it doesn't do the user or database creation step. Would this be sufficient for this functionality or would it still be necessary to skip  CreateUserDatabase completely?

src/postgresql/pgolap.tcl
sm-shaw: Yes, any options for the CLI should be included in the GUI options as well. This requires updating pgopt.tcl to add an additional row to the interface. 
Also any options added should be included in the connectpool options as well.

src/postgresql/pgolap.tcl
sm-shaw: requiressl is a deprecated option in favour of sslmode along with an number of additional SSL options. 
All options can be seen with:
`foreach prop [ pg_conndefaults ] { puts $prop } `
If SSL support is required for PostgreSQL then I would suggest that this be a separate task and not tied to Azure in particular. If this is needed I'd suggest adding an Issue for adding SSL Support to PostgreSQL as a starting point. There are connections in multipl places include the standard build and driver script but also the asynchronous connections and connect pooling and any ssl option would be to be selectable in both CLI and GUI.

config/postgresql.xml
```diff
@@ -3,6 +3,7 @@
     <connection>
         <pg_host>localhost</pg_host>
         <pg_port>5432</pg_port>
+        <pg_azure>false</pg_azure>
```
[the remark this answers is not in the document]
raymond13513: Flexible server is in preview right now. We want to support benchmarking databases that are in production. Since we have is_azure in SQL server I followed the naming convention. Would is_azure_single_server a better name? 

I'm not sure if removing CreateUserDatabase will affect the benchmark or not

src/postgresql/pgolap.tcl
raymond13513: Will look into it

sm-shaw:
As general feedback there seems to be a couple of tasks here and it would be helpful to break them out into 2 issues/pull requests.  Firstly if there is a requirement for PostgreSQL to support SSL then this would not necessarily be specific to Azure and therefore should not be enabled under an Azure only option but be a separate Issue and then pull request created to add a general SSL option that would activiate the sslmode option in the PgTcl client when selected.  Note that connections are made in the standard driver scripts as well as the asynchronous and connect pooling options. The asynchronous and connect pooling options modify the standalone scripts when selected. Any changes need to go into the GUI (e.g. a checkbox named enable SSL) and the CLI. 

Secondly there is the issue of compatibility with Azure single server when the user contains @machine_name. It appears that there could be a simpler way to do this. I cannot test Azure directly myself at present however @marcocitus does say that by commenting out CreateUserDatabase, Azure Single Server works fine.  This is only in the schema build so it is not entirely clear if @machine_name is needed in the driver script as well? In the next version CreateUserDatabase is modified to skip creation if the user and database exists, however it would be not be too difficult to also add an option to skip this entirely if it was wanted to do everything manually before the schema build. This would also have the advantage of not being Azure specific. The more generic the code can be for multiple use cases the better.

raymond13513:
Ok, I'll split this into two issue

src/postgresql/pgoltp.tcl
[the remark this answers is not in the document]
marcocitus: There's a general problem that managed PostgreSQL services do not provide superuser access, so when using HammerDB you have no choice but to pass non-superuser details into the superuser fields. Usually, the initial database user generated by the managed service does have many privileges (e.g. creating or altering a database).

As long as the privileged user has permission to run all the commands in CreateUserDatabase, or at least  to run the commands that are used when the database/user already exist (after #132), HammerDB works. It is unclear whether this is true for all managed PostgreSQL services.

The problem with the Azure single server is two-fold:

1) In the connection string the username has to include @machine_name, which is used by an Azure load balancer to identify the database server. On the database server itself, the username does not include the machine name. The current implementation of HammerDB quite reasonably assumes the username used in the connection string to also be usable as the username inside the database for the purpose of CREATE ROLE and GRANT commands, but this is not the case. 

2) Even after this PR, which attempts to address the first point by stripping the @machine_name part from the username in CreateUserDatabase, I get the following error from the ALTER USER .. PASSWORD .. command in CreateUserDatabase.

```sql 
ALTER USER marco PASSWORD '...'
ERROR:  must be superuser to alter replication users
```

So it seems that Azure single server does not allow all of the steps in CreateUserDatabase. 

Perhaps this specific situation could be addressed, but overall it seems that an option to skip CreateUserDatabase would allow users to always avoid managed service idiosyncrasies, because the remainder of the benchmark only needs permission to create tables and functions.

marcocitus:
I get a syntax error when using the current version:
```sql
Error in Virtual User 1: ERROR:  syntax error at or near "@"                                                                                                                      
LINE 1: ALTER USER marco@marco-machine PASSWORD '...'
```

After fixing that, I still get the following error:
```sql
Error in Virtual User 1: ERROR:  must be superuser to alter replication users   
```

src/postgresql/pgoltp.tcl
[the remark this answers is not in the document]
sm-shaw: Marco, many thanks for the detailed update and explanation. So it looks clear that adding an option to skip CreateUserDatabase in the schema builds for both the TPROC-C and TPROC-H workloads (in both GUI and CLI) is the best option to solve this issue. This also keeps compatibility with non managed service implementations as it is then possible to test in a standard environment. 
This should also be distinct from any use of SSL for connections.

Issue #168, opened 2020-11-20
distributescript results in invalid command name "applyctexthighlight"


sm-shaw:
When using Primary/Replica mode, distributing the script from the Primary to the Replica results in the error: invalid command name "applyctexthighlight" on the replica. 
```
hammerdb>Script loaded, Type "print script" to view
Error: invalid command name "applyctexthighlight"
```
applyctexthighlight is a GUI function only. As a workaround the following can be run at the command line on the replica or added to the script to be run on the replica to prevent the error.
`
hammerdb>proc applyctexthighlight { args } { ; }`

PR #169, opened 2020-11-20
Fix distributescript applyctexthighlight error

sm-shaw:
Fix for Issue #168


abondvt89:

Issue #164, opened 2020-10-29
Postgres TPCH are missing Indexes


raymond13513:
The Postgres TPCH benchmark are timing out after the first three queries. 
After a bit digging, I found that Indexes are not created when running Postgres TPCH. 

Here is another article in Japanese that talk about this issue.
https://atsuizo.hatenadiary.jp/entry/2019/09/04/090000

After adding indexes to the tables. The TPCH benchmark becomes comparable to other database engine.
CREATE INDEX ON customer (C_NATIONKEY);
CREATE INDEX ON lineitem (L_ORDERKEY);
CREATE INDEX ON lineitem (L_SUPPKEY);
CREATE INDEX ON lineitem (L_PARTKEY,L_SUPPKEY);
CREATE INDEX ON nation (N_REGIONKEY);
CREATE INDEX ON orders (O_CUSTKEY);
CREATE INDEX ON partsupp (PS_PARTKEY);
CREATE INDEX ON partsupp (PS_SUPPKEY);
CREATE INDEX ON supplier (S_NATIONKEY);
CREATE INDEX ON lineitem (L_SHIPDATE);
CREATE INDEX ON lineitem (L_COMMITDATE);
CREATE INDEX ON lineitem (L_RECEIPTDATE);
CREATE INDEX ON orders (O_ORDERDATE);

sm-shaw:
Many thanks for the Issue. From a quick check and comparison with Oracle the same indexes are created so it is not the case that Indexes are not created at all.  However it may be the case that additional indexes could benefit performance so we will take a look into this as any changes needs to work for multiple schema sizes and also work with the refresh function. 

PostgreSQL
```
schemaname | tablename |  indexname  | tablespace |                                         indexdef                                          
------------+-----------+-------------+------------+-------------------------------------------------------------------------------------------
 public     | region    | region_pk   |            | CREATE UNIQUE INDEX region_pk ON public.region USING btree (r_regionkey)
 public     | nation    | nation_pk   |            | CREATE UNIQUE INDEX nation_pk ON public.nation USING btree (n_nationkey)
 public     | supplier  | supplier_pk |            | CREATE UNIQUE INDEX supplier_pk ON public.supplier USING btree (s_suppkey)
 public     | partsupp  | partsupp_pk |            | CREATE UNIQUE INDEX partsupp_pk ON public.partsupp USING btree (ps_partkey, ps_suppkey)
[...]

PR #170, opened 2020-11-20
Add indexes to PostgreSQL Foreign Keys

sm-shaw:
Add indexes to Foreign Keys as per Issue #164


abondvt89:

Issue #103, opened 2020-02-28
Enable Variable Load/Stress Testing


sm-shaw:
Enable the Load of a HammerDB OLTP/TPC-C test to be varied throughout the test. Currently HammerDB tests have a defined number of virtual users that are fixed throughout the duration of the workload.  Add the option to vary the load throughout the test. The maximum number of virtual users would be created and fixed at the start of the test however a number of variables given at the start of the test could vary the load. a) the time the virtual user starts running. b) the number of asynchronous client managed by each virtual user with keying and thinking time. c) the keying and thinking times of each thread. – these would be user configurable. The goal of this enhancement is to test database environments that respond to increased demand through increased resources (elasticity), in particular measuring the transition periods. 
-	Number of virtual users.
-	Starting number of connections (per VU)
-	Number of seconds and increase in connections (per VU) i.e. the step profile.
-	Number of steps to be performed.
-	Option to reverse once the number of steps have been completed i.e. reduce the number of connections by the interval,
-	Option to create all connections immediately but only send calls down them according to the step function.
-	Keying and thinking time, as currently implemented (so no change required)

![image](https://user-images.githubusercontent.com/38044085/75543975-ebcdbd00-5a1a-11ea-996d-95a40ee1b644.png)

Additional requirements would be more fine grained error handling and reporting for the duration of the test. We should have the option to log EVERY error, with the timestamp.  In particular transaction response time measurements would need to be taken for multiple virtual users. Currently etprof uses a high resolution timer requiring investigation of alternatives. Key focus is on data for transition periods. This data could be captured into SQLite for reporting at the end of the run.  We could choose to increase the frequency of response time collection immediately after a transition, governed by “every n seconds for m seconds after each transition”.

Stress testing follows a similar approach however the goal is to test the sensitivity of the database to the oversubscription of connections. The clients using the TPC-C schema execute an insert followed by select statement without using the stored procedures. The load is continually increased. The maximum number of virtual users is defined in advance.  Once a threshold of either errors or response times (measured by the monitor virtual user) or all virtual users are running is reached the test is terminated, rather than being time based.

sm-shaw:
Feature in PR #211 adds step workload feature enabled by automatically creating primary and replicas activated by steprun command. Feature will be CLI only as multiples replicas are created. Replicas are defined in steps.xml in config directory.
```
<steps>
	<replica1>
		<start_after_prev>1</start_after_prev>
		<duration>4</duration>
[...]

PR #178, opened 2020-12-02
Add task manager module

sm-shaw:
Add task manager module for better command line timing based on coroutines. This module enables the creation of step/variable workloads as described in #103 and allows out to the main thread during driver script runs improving on runtimer.


abondvt89:

Issue #179, opened 2020-12-03
Improve CLI runtimer by making asynchronous


sm-shaw:
HammerDB v4.0 includes a runtimer procedure that keeps the main thread busy while virtual users are running a workload in a sequence. An example of such a script is shown below:
```
#!/bin/tclsh
puts "SETTING CONFIGURATION"
dbset db pg
dbset bm TPC-C
diset tpcc pg_superuser postgres
diset tpcc pg_superuserpass postgres
diset tpcc pg_user tpcc
diset tpcc pg_pass tpcc
diset tpcc pg_total_iterations 10000000
diset tpcc pg_driver timed
diset tpcc pg_rampup 0
diset tpcc pg_timeprofile false
diset tpcc pg_duration 2
vuset logtotemp 1
loadscript
puts "SEQUENCE STARTED"
foreach z { 1 2 3 } {
puts "$z VU TEST"
vuset vu $z
vucreate
vurun
runtimer 150
vudestroy
}
after 5000
puts "TEST SEQUENCE COMPLETE"
```
This runtimer procedure times a synchronous delay for 1 second. Typcially no output is expected and therefore there is no impact on the timing. 
```
CURRENT RUNTIMER
Vuser 1:1 Active Virtual Users configured
Vuser 1:TEST RESULT : System achieved 46892 NOPM from 108079 PostgreSQL TPM

ASYNCH RUNTIMER
Vuser 1:1 Active Virtual Users configured
Vuser 1:TEST RESULT : System achieved 46120 NOPM from 105984 PostgreSQL TPM
```
However if a timeprofile is run such as:
`diset tpcc pg_timeprofile true`

Then the main thread will block receiving the percentile output for second intervals. This does not impact the timings themselves which are gathered before sending output, however will reduce the transaction rate in the first active virtual user only while it is waiting to send the output to the main thread.  The example shows the current synchronous versus an asynchronous time for one virtual user only. This does not affect any other virtual users from the first active one as only the first active one is doing the timing. 
```
CURRENT RUNTIMER TIMEPROFILE
Vuser 1:TEST RESULT : System achieved 9973 NOPM from 22760 PostgreSQL TPM
Vuser 2:+-----------------+--------------+------+--------+--------------+--------------+
Vuser 2:|PROCNAME | EXCLUSIVETOT| %| CALLNUM| AVGPERCALL| CUMULTOT|
Vuser 2:+-----------------+--------------+------+--------+--------------+--------------+
[...]

PR #180, opened 2020-12-03
Update CLI runtimer to asynchronous

sm-shaw:
Change runtimer from synchronous (with update every second) to asynchronous as per #179


abondvt89:

Issue #187, opened 2021-01-08
Update PostgreSQL Data Types and Hash Partition


sm-shaw:
Blog post from EnterpriseDB here https://www.enterprisedb.com/blog/postgresql-tproc-c-benchmarks-postgresql-12-vs-postgresql-13-performance suggests changing Data Types for the Schema. Trying the suggested changes brings 2.7% performance benefit at peak. Hash Partitioning Order Line is a common technique for other databases now supported in PostgreSQL, this brings an additional 0.3% making 3% performance gain in total from applying these changes according to ISV recommendations.

PR #188, opened 2021-01-08
 Update PostgreSQL Data Types and Hash Partition for Issue #187

sm-shaw:
Updates as per Issue #187, also required to update the calling of the ostat function to align with the correct data types.


abondvt89:
Accepting pull request as voted on January 19, 2021 in the TPC-OSS subcommittee.

Issue #189, opened 2021-01-13
Oracle 19C Incorrect AWR Snapshots Reported in Multitenant DB


dfhawthorne:
When run a TPC-C workload against an Oracle 19C PDB, the AWR snapshots reported are for the CDB not the PDB:
```
Vuser 1:Start Snapshot 220 taken at 13 JAN 2021 17:31 of instance DEMO19C (1) of database DEMOCDB1 (1856558291)
Vuser 1:Timing test period of 5 in minutes
Timer: 3 minutes elapsed
Vuser 1:1 ...,
Timer: 4 minutes elapsed
Vuser 1:2 ...,
Timer: 5 minutes elapsed
Vuser 1:3 ...,
Vuser 1:4 ...,
Vuser 1:5 ...,
Vuser 1:Test complete, Taking end AWR snapshot.
Vuser 1:End Snapshot 220 taken at 13 JAN 2021 17:31 of instance DEMO19C (1) of database DEMOCDB1 (1856558291)
Vuser 1:Test complete: view report from SNAPID 220 to 220
Vuser 1:2 Active Virtual Users configured
Vuser 1:TEST RESULT : System achieved 195 NOPM from 0 Oracle TPM
```
__Note__ The same SNAPID is reported as the start and end values.

This issue appears to be similar to #13.

sm-shaw:
Thanks for the Issue and proposed solution. I've tested it on 12c and confirmed it doesn't have an impact on the existing version. Currently creating a pluggable 19c install to test there so may take a little while to test and give feedback.

sm-shaw:
So I setup a 19c/19.3 database with a pluggable database.  The 12c version I had used was 12.1 and the behaviour does seem to be different. It has been a while since I looked into this so I might easily be missing something. To test I used the unmodified HammerDB v4.0 without the patch.  I set up a container database RAZCDB1 and pluggable RAZPDB1. When I installed the test schema in RAZPDB1 and ran a test it appeared to take the snapshots from the container and also reported the TPM. 

```
Hammerdb Log @ Fri Jan 15 10:26:09 EST 2021
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-
Vuser 1:Beginning rampup time of 1 minutes
Vuser 2:Processing 1000000 transactions with output suppressed...
Vuser 3:Processing 1000000 transactions with output suppressed...
Vuser 1:Rampup 1 minutes complete ...
[...]

PR #190, opened 2021-01-13
Proposed solution to #189 - Wrong AWR Snapshots listed

dfhawthorne:
This is my proposed solution to issue #189 .

The code has only been tested on an Oracle 19C PDB. The fix is only applied if the DB major version is greater than or equal to 19. 

After applying the fix, I get the correct AWR snapshots listed:
```
Vuser 1:Rampup complete, Taking start AWR snapshot.
Vuser 1:Start Snapshot 94 taken at 13 JAN 2021 22:05 of instance DEMO19C (1) of database DEMOCDB1 (908248820)
Vuser 1:Timing test period of 5 in minutes
Vuser 1:1 ...,
Vuser 1:2 ...,
Vuser 1:3 ...,
Vuser 1:4 ...,
Vuser 1:5 ...,
Vuser 1:Test complete, Taking end AWR snapshot.
Vuser 1:End Snapshot 95 taken at 13 JAN 2021 22:10 of instance DEMO19C (1) of database DEMOCDB1 (908248820)
Vuser 1:Test complete: view report from SNAPID 94 to 95
```


sm-shaw:
comments are in Issue #189, In testing firstly the unmodified version against a pluggable database in Oracle 19.3 it correctly reported the TPM and also gave the option of "AWR_PDB - Use AWR data from PDB" to select the snapshot so appeared to work correctly in this version without needing the modification. Feedback is needed before proceeding as it is not clear if this fix is now needed as it looks as if Oracle have made changes to how pluggable databases work in 19c.

abort:
Hi, I have tested the latest origin/master commit as well as this PR on an Oracle EE 19.3 (Windows) DB. I can confirm that this PR was required for me in order to get the TPM numbers to show!

sm-shaw:
Thanks for the additional info - I think I need to go back and check what is different in my recent 19.3 installation as the output in #189 showed that the pluggable database seemed to work OK with v4.0.

dfhawthorne:
I have fixed a bug with my original PR in that the fix did not work for connections to the root container.

sm-shaw:
Please see further updates in #189.  Still unable to replicate the behaviour that the PR fixes in PDBs created in 19.3 on Linux. Clearly this prevents the PR being accepted until it is possible to reproduce.  For example can you provide the output from: 
```
SQL> select text_vc from dba_views where view_name = 'DBA_HIST_SNAPSHOT'
SQL> select text_vc from dba_views where view_name = 'DBA_HIST_SYSSTAT'
```
to verify that these views are the same as the output I have pasted.

dfhawthorne:
My output was posted in #189. I cannot see any difference in the view definitions. We appear to be at different patch levels in the database software (19.3 vs 19.9).

sm-shaw:
additional information in #189 - Oracle Doc ID 2711623.1 says that this functionality (to query DBA_HIST_SNAPSHOT/DBA_HIST_SYSSTAT as HammerDB does) from a PDB was included from 18.1. This is what testing has shown. Therefore this change should not be required to support 19c.

dfhawthorne:
Made changes as suggested. Quoted wrong issue (#188) instead of #189 in commit message.

sm-shaw:
This looks great, it also needs changing for exactly the same statement further down. There are 2 timed scripts one for timed and then one for asynchronous timed. If you search for the comment #ASYNCHRONOUS TIMED SCRIPT then the unchanged statement will be following this. So browsing the changes we will see the same change twice in the file. With this additional commit then it will be good to accept.

dfhawthorne:
Made change for ASYNC as suggested. Test results posted in #189

PR #193, opened 2021-02-02
Swap column order in stock index in pgoltp.tcl

SaitTalhaNisanci:
It seems like https://github.com/TPC-Council/HammerDB/commit/3d63e3e4e7b4d399a37cfec636e6839af85bca85#diff-5fbe626126351cf83ebe51980431f33115232da6c5ce75e402e331a55e279e9aR1663 reverted the fix done in https://github.com/TPC-Council/HammerDB/pull/124 by accident. This PR fixes that.


sm-shaw:
Many thanks for picking this up. Yes this change was inadvertent and not picked up during testing so thanks again for the correction to restore the changes from #124

Issue #194, opened 2021-02-15
 Postgres datatype change from number to integer performance degragation 


greenhal:
I noticed a drop in performance with the table column datatype change from number to integer, introduced in commit 9b0305e3eab18af589aad6c5bf219bd7d011318f (1/26/2021).

Performance dropped about 30%.    I was able to trace to the payment procedure, 
Specifically the INOUT parameter p_c_id.

When defined as a NUMERIC(5,0),  the query plan does does not use the full index (c_w_id, c_d_id , c_id), it does scan on c_w_id, c_d_id and filter on c_id,

When defined as a INTEGER, the query plan does use the full index,

_**NUMERIC query plan:**_

2021-02-12 18:27:27.993 UTC [16541] LOG: duration: 2.390 ms plan:
Query Text: SELECT c_first, c_middle, c_last, ,,,
FROM customer
WHERE c_w_id = p_c_w_id AND c_d_id = p_c_d_id AND c_id = p_c_id
**Index Scan using customer_i1 on customer (cost=0.43..5738.56 rows=15 width=130) (actual time=2.388..2.388 rows=1 loops=1)
*Index Cond: ((c_w_id = 28) AND (c_d_id = 3))*
*Filter: ((c_id)::numeric = '2998'::numeric)*
Rows Removed by Filter: 2997**
2021-02-12 18:27:27.993 UTC [16541] CONTEXT: SQL statement "SELECT c_first, c_middle, c_last ,,,
FROM customer
WHERE c_w_id = p_c_w_id AND c_d_id = p_c_d_id AND c_id = p_c_id“
PL/pgSQL function payment(integer,integer,integer,integer,integer,numeric,character,.. ) line 41 at SQL statement

_**INTEGER query plan:**_

2021-02-12 19:05:16.670 UTC [19065] LOG: duration: 0.009 ms plan:
Query Text: SELECT c_first, c_middle, c_last, ...
FROM customer
WHERE c_w_id = p_c_w_id AND c_d_id = p_c_d_id AND c_id = p_c_id
**Index Scan using customer_i1 on customer (cost=0.43..8.45 rows=1 width=130) (actual time=0.008..0.008 rows=1 loops=1)
*Index Cond: ((c_w_id = 47) AND (c_d_id = 1) AND (c_id = 746))***
2021-02-12 19:05:16.670 UTC [19065] CONTEXT: SQL statement "SELECT c_first, c_middle, c_last, ...
FROM customer
FROM customer
[...]

PR #204, opened 2021-02-22
Fix for Issue #194 Postgres datatype change from number to integer

sm-shaw:
Fix for Issue #194 where datatype change caused a performance regression when the stored procedure option is selected. Update has been tested and performance restored changing numeric to integer for the payment stored procedure. Other stored procedures also checked and no further similar issues found.


sm-shaw:
Accepting pull request as voted on March 2, 2021 in the TPC-OSS subcommittee.

Issue #216, opened 2021-03-17
Error when running TPC-C MySQL with XML Connection Pool in HammerDB v4.0


atsuizo:
Could you inform me what settings to use to avoid this error ?

And if there is no documentation on how to set that up, It would be very helpful if you could add it.

I really wanted to contribute by finding ways to cope with the situation, but I didn't have the skills to do so.

## Error Messages

```
mysqlexec/db server: Unknown prepared statement handler (ostat_st) given to EXECUTE
mysqlexec/db server: Unknown prepared statement handler (delivery_st) given to EXECUTE
```
MySQL Error Code 1243.

This message is displayed for all VUs.

## Settings

・dict
```
mysql_connect_pool     = true
```
For the other parameters, I used the same values as the settings that work fine in non-ConectionPool mode.
(Even if I don't set "mysql_prepared = true", it will be set to true internally in ConnectionPool mode, right?)

・mysqlcpool.xml
```
・Connections
c1：Master Instance Endpoint & Correct Port, User, Password
c2：Read Replica Instance Endpoint & Correct Port, User, Password

・sprocs
<neworder>
        <connections>c1</connections>
        <policy>round_robin</policy>
</neworder>
<payment>
        <connections>c1</connections>
        <policy>round_robin</policy>
</payment>
<delivery>
        <connections>c1</connections>
        <policy>round_robin</policy>
</delivery>
<stocklevel>
        <connections>c2</connections>
        <policy>round_robin</policy>
</stocklevel>
<orderstatus>
        <connections>c2</connections>
        <policy>round_robin</policy>
</orderstatus>
```

I installed HammerDB v4.0 on AWS EC2 - Amazon Linux2,  and set up a MySQL 8.0.23 client using "yum".

I tried it with AWS RDS/MySQL 8.0.21 and AWS Aurora MySQL 5.7（mysql_aurora.2.09.2）, reproduced it.

If in ConectionPool mode also directs all "sprocs.connections" to C1(Master Instance) only, this message will not be output.

If we set C2 only for "sprocs.orderstatusconnections" and C1 for the rest, an error message of "slev_st" will be output instead of  "delivery_st" .

Furthermore, when I did the equivalent setting in PostgreSQL, no error message was output.

sm-shaw:
Many thanks for the Issue, yes you have found a bug. There is a debug line that if uncommented prints out the prepared statements.
```
[...]

PR #217, opened 2021-03-17
Fix for Issue #216 Unkown prepared statement handler with connect pool

sm-shaw:
Fix for Issue #216 by reordering prepared statements


abondvt89:
Accepting pull request as voted on March 23, 2021 in the TPC-OSS subcommittee.

Issue #224, opened 2021-04-01
Extended Time Profiling


sm-shaw:
HammerDB currently uses the etprof package for time profiling.  Due to the profiling overhead this package is restricted to running in the first active virtual user only. This does not give a full picture on all of the timings across the workload. 
The requirement is to build an alternative time profiling package that can be loaded to measure the time profiles for all virtual users more efficiently and generate the output to a separate logfile. It will be optional whether to use extended time profiling or to use etprof to record percentiles at 10 second intervals for the first active virtual user.

sm-shaw:
Solution adds a new time profile section to generic.xml.
```
<timeprofile>
           <profiler>xtprof</profiler>
           <xt_unique_log_name>0</xt_unique_log_name>
    </timeprofile>
```
User continues to choose a time profile option or checkbox as before, behaviour is determined by xml options. 
If profiler is set to etprof time profiling works as before measuring timings in Virtual User 2 only. 
If profiler is set to xtprof it will capture timings for all virtual users and report them at the end including a summary of all timings. Output is to file hdbxtprofile.log, the only current option is to create a unique log file name. All timing data is collated by the monitor virtual user.

Timing Data Written is:

CALLS: Number of calls to the stored procedure through the timing period
MIN: Minimum response time
AVG: Average response time
MAX: Max response time
TOTAL: Total response (for the summary this will be more than elapsed time as a sum of all virtual users)
P99: 99th percentile
P95: 95th percentile
P50:50th percentile
SD:Standard Deviation
RATIO: Percentage of calls by that Virtual User - output is ordered by ratio from high to low

Timing profile overhead measured on a latest generation server was 4%

Example below for SQL Server on test PC
```
MSSQLServer Hammerdb Time Profile Report @ Fri Apr 02 10:23:56 BST 2021
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
>>>>> VIRTUAL USER 2 : ELAPSED TIME : 421062ms
>>>>> PROC: NEWORD
CALLS: 169244	MIN: 0.430ms	AVG: 0.994ms	MAX: 1042.998ms	TOTAL: 168189.754ms
P99: 2.465ms	P95: 1.383ms	P50: 0.857ms	SD: 41300.051	RATIO: 39.944%
>>>>> PROC: PAYMENT
[...]

PR #225, opened 2021-04-02
Extended Time Profiling as per Issue #224

sm-shaw:
Pull Request implements extended time profiling that captures timings for all Virtual Users and writes output to logfile.


abondvt89:
Accepting pull request as voted on April 5, 2021 in the TPC-OSS subcommittee.

PR #228, opened 2021-04-08
Update ora_timeprofile variable for runstep compatibility

sm-shaw:
fix error in steprun command for oracle with introduced mismatch from timeprofile variable


sm-shaw:
minor bugfix in oracle steprun

Issue #250, opened 2021-05-28
Increase TPROC-C Schema Size Limits


sm-shaw:
HammerDB limits the schema build size in the GUI/CLI to 5,000 warehouses and 30,000 with datagen. These limits are not restrictions within the build script itself but are instead set prior to the build to prevent oversizing. It has been requested to increase this limit as follows:

1. Update datagen limit of 30000 to 100000
2. Update GUI/CLI build limit of 5000 to 100000
3. Additionally, increase number of default transactions from 1000000 to 10000000

To test schema/stored procedure changes it is possible to modify the build/driver scripts to start at the new upper limit instead of needing to build up to 100K WH builds to test all databases.
In the build script the start and end can be modified as follows:
```
set remb [ lassign [ findchunk $num_vu $count_ware $myposition ] chunk mystart myend ]
set mystart [ expr $mystart + 100000 ]
set myend [ expr  $myend + 100000 ]
```
and in the driver script
`set w_id  [ RandomNumber 100001 $w_id_input ]  `

A test of all support databases shows the following summary:
Db2,  SQL Server do not require schema changes
MySQL, MariaDB also work without errors as limits are display limits instead of precision but should nevertheless be updated accordingly
PostgreSQL requires updates to the NEWORD function and stored procedure
Oracle needs updates to X_W_ID fields

A more detailed summary of tests as follows:
Db2
Needed schema updates as follows to check the build. Check showed that build and test with 100K W_ID then worked without further update. Schema change as below will not be needed as w_id will always start from 1 and end at $count_ware. Change is only for testing the build to prevent deadlock errors. Test showed that Db2 build and test did not need modification.
```
CREATE TABLE WAREHOUSE (W_NAME CHAR(10), W_STREET_1 CHAR(20), W_STREET_2 CHAR(20), W_CITY CHAR(20), W_STATE CHAR(2), W_ZIP CHAR(9), W_TAX REAL, W_YTD DECIMAL(12, 2), W_ID INTEGER NOT NULL) IN [ dict get $tspace_dict W ] INDEX IN [ dict get $tspace_dict W ] ORGANIZE BY KEY SEQUENCE ( W_ID STARTING FROM 100001 ENDING AT [ expr $count_ware + 100000 ]  ) ALLOW OVERFLOW"
```
SQL Server - No modifications needed
Tested Standard Schema, In-Memory OLTP  and Stored Procedures

PostgreSQL
Standard Schema build OK
Functions:
```
[...]

PR #251, opened 2021-06-14
Update to increase TPROC-C Schema Size Limits for Issue #250

sm-shaw:
Increase soft warehouse limit from 5000 GUI/30000 Datagen to 100000 for both.
Also increase default transactions per virtual user from 1000000 to 10000000.


abondvt89:
Accepting pull request as voted on June 16, 2021 in the TPC-OSS subcommittee.

PR #252, opened 2021-06-15
Update versions and changelog for v4.2

sm-shaw:
Updates version number and changelog for v4.2


abondvt89:
Accepting pull request as voted on June 16, 2021 in the TPC-OSS subcommittee.

PR #254, opened 2021-06-18
Fix for Xtprof can retain old data

sm-shaw:
Reinitialize xtprof thread shared variables on module load.


abondvt89:
Accepting pull request as voted on June 29, 2021 in the TPC-OSS subcommittee.

Issue #264, opened 2021-07-26
[TPROC-H] MySQL create FKs after table load.


lrweck:
Creating the FKs beforehand makes loading large amounts of data unnecessarily slow, as it could (and should) be created afterwards, like it does on PostgreSQL.

sm-shaw:
Can you provide more information on your test environment, in particular the storage engine you are using. Although the foreign keys are included in the table definitions, the default storage engine is MyISAM and when this was checked most recently queries on MyISAM were still faster than InnoDB.  Also, have you tried the system variable foreign_key_checks, setting this to 0 should allow the foreign keys to be created but not checked during the build.

lrweck:
Hi! I'm using InnoDB.

My objective was to create similar environments for testing both MySQL and PostgreSQL. I know loading data is not supposed to be used to be a benchmark, but since the code is similar and we're talking about the same data being loaded, it could reasonably be something measured. I believe simply disabling foreign_key_checks still incurs a penalty for each INSERT, so it's not really the same as creating the index from scratch.

So I'd like to propose having the same loading methods for at least PostgreSQL and MySQL (and MariaDB, since it's basically MySQL)

sm-shaw:
Have you tried setting foreign_key_checks to 0 at a system level? 
The simplest solution to account for different storage engines would be to set foreign_key_checks to 0 within the build script, this would be straightforward. 
I would recommend some caution about expecting exactly the same performance from MariaDB and MySQL they have diverged somewhat and therefore this shouldn't necessarily be assumed depending on the versions you are testing.

lrweck:
I haven't tried setting it... However, FK checking is just part of the question. Even if they are disabled, each insert still has to update their respective indices.

sm-shaw:
Thanks, when time permits we'll benchmark a number of options to evaluate the best approach.

lrweck:
Actually, HammerDB already sets "foreign_key_checks = 0"... 
![image](https://user-images.githubusercontent.com/26086198/127578003-14515c26-3b0f-48db-9d2f-a97bfc708ade.png)

sm-shaw:
Good catch.  If you wish to submit a pull request to create the indexes and foreign keys after the build and can show the performance improvement for different storage engines then that would be a good approach.  Otherwise this will be added to the list of development task.

lrweck:
[...]

PR #274, opened 2021-08-27
Update for #264 [TPROC-H] MySQL create FKs after table load.

sm-shaw:
Updates after tests for Issue #264 - creating foreign keys after table load did not prove effective, however investigation showed that existing indexes were not compatible with TPC-H specification and therefore build updated for compliance. Default storage engine changed to innodb as updated configuration performed better than previous default myisam. Updates for both MySQL and MariaDB. 
Test scripts attached to Issue #264 - potential to revisit configuration and build if MySQL/MariaDB supports features such as parallel index builds and/or parallel query in the future.


abondvt89:
Accepting pull request as voted on September 7, 2021 in the TPC-OSS subcommittee.

PR #283, opened 2021-11-09
v4.3 fix - encapsulate graphical metrics in namespace

sm-shaw:
Testing showed some function name clashes in graphical metrics with PostgreSQL and Oracle. To make the solution cleaner, this pull request creates a namespace for both PostgreSQL and Oracle metrics, exports all of the function name and imports the required one and forgets the unneeded one on loading. 
Also includes a minor change of making default SQLite file for web service on disk in temp area rather than in memory. in-memory does not allow connection from multiple threads preventing additional functionality such as timing. Very minor change in the XML config.


abondvt89:
Accepting pull request as voted on November 10, 2021 in the TPC-OSS subcommittee.

PR #298, opened 2021-12-15
Make Linux hammerdb executable work with spaces in PATH

JelteF:
If the $PATH variable contained spaces then it would result in an error
like this:

```
./hammerdb: 6: export: (x86)/Microsoft: bad variable name
```

By double quoting the usage of the existing $PATH variable bash handles
this correctly.

### DIFF
--- a/hammerdb
+++ b/hammerdb
@@ -1,9 +1,9 @@
 #!/bin/sh
 ########################################################################
 # \
-export LD_LIBRARY_PATH=./lib:$LD_LIBRARY_PATH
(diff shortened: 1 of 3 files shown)


sm-shaw:
I have tested and confirmed that the error "bad variable name" is received if there are spaces already in the PATH environment variable, and this is resolved by double-quoting. I have also tested on Windows to verify there is no impact. 
This is all fine, so recommended for acceptance by the TPC-OSS subcommittee.

PR #301, opened 2021-12-17
Build dataset using COPY instead of multi-row inserts for TPROC-C

JelteF:
NOTE: this PR includes the #292, to make it more readable and make merging them
both in the future easier. Adding comments to the diff is probably easier to do
on this equivalent PR with a different base branch on the Citus repo: https://github.com/citusdata/HammerDB/pull/3

The fastest way to bulk insert data in Postgres is by using COPY. This 
changes the dataset building to use that for TPROC-C. I tried building a dataset for 
1000 warehouses using 100 vusers. Without copy this took 100 minutes, with
copy it only took 42 minutes. So it reduced the time it takes to build the
dataset by ~58%.

Docs on the usage of COPY in the tcl Postgres library can be found here: 
http://pgtclng.sourceforge.net/pgtcldocs/pgtcl-example-copy.html

NOTE: A similar improvement could probably be done for postgres on TPROC-H

### DIFF
--- a/modules/tpcccommon-1.0.tm
+++ b/modules/tpcccommon-1.0.tm
@@ -1,315 +1,315 @@
 package provide tpcccommon 1.0
 namespace eval tpcccommon {
-namespace export chk_thread RandomNumber NURand Lastname MakeAlphaString Makezip MakeAddress MakeNumberString findchunk findvuposition randname keytime thinktime async_keytime async_thinktime async_time get_connect_xml
-#gettimestamp not included as uses different formats per database
-#TPCC BUILD PROCEDURES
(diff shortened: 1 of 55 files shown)


sm-shaw:
I have tested this change on the same system with the changes for #295 already in place. 
The schema build is for TPROC-C 800WH with 64VUs
for #295 the results were:
postgres orig = 12 min 40 secs / postgres new = 8 min 54 secs = 30% improvement
With #301 and #295 together, the time to build the schema was 5 min 57 secs = 53% improvement on original build time.
PR is recommended for approval.

sm-shaw:

PR #303, opened 2021-12-22
readline: only catch signals when using interact

JelteF:
When running `./hammerdbcli auto somescript.ctl` on Linux it was
impossible to stop the benchmark by using Ctrl+C. This was pretty
annoying since it is the most common way to stop programs in a shell on
Linux. The reason was that the readline module was explicitly ignoring
SIGINT. This makes sense for interactive mode, but when using `auto`
there's no reason for this.

This change makes sure the signal capturing code is only executed once
`TclReadLine::interactive` is called.


sm-shaw:

Issue #289, opened 2021-12-08
getcols can return zero, resulting in divide-by-zero


dannysauer:
TLDR: It would probably be good if this code either fell back to a sane default of 25x80 or accepted the `$COLUMNS` environment variable (or both), preventing mysterious failures in edge cases. :)  On to the details:

https://github.com/TPC-Council/HammerDB/blob/a86e7f066c3ef40354258dd6e7432308a3fd7044/modules/tclreadline-1.2.tm#L129-L152

I get that the docker container isn't an official image, but when run in Kubernetes interactive mode, getcols can return a terminal width of zero.  That ultimately results in an experience like this:
```
sauer@GRODD:~> kubectl -n danny-yugabyte-demo run --rm -i --tty --restart=Never --image webysther/hammerdb hammerdb-danny -- ./hammerdbcli
HammerDB CLI v3.3
Copyright (C) 2003-2019 Steve Shaw
Type "help" for a list of commands
The xml is well-formed, applying configuration
divide by zero
    while executing
"expr {$cursorLen % $COLUMNS}"
    (procedure "prompt" line 38)
    invoked from within
"prompt $CMDLINE"
    (procedure "tclline" line 125)
    invoked from within
"tclline "
    (procedure "TclReadLine::interact" line 41)
    invoked from within
"TclReadLine::interact"
    invoked from within
"if { $autostart::autostartap == "true" } {
	source $autostart::autoloadscript
	} else {
    TclReadLine::interact
	}"
    (file "./hammerdbcli" line 84)
pod "hammerdb-danny" deleted
pod danny-yugabyte-demo/hammerdb-danny terminated (Error)
```

Running the same command, but executing bash and then running hammerdbcli is perfectly happy:
```
sauer@GRODD:~> kubectl -n danny-yugabyte-demo run --rm -i --tty --restart=Never --image webysther/hammerdb hammerdb-danny -- /bin/bash
If you don't see a command prompt, try pressing enter.
root@hammerdb-danny:/hammerdb# ./hammerdbcli 
HammerDB CLI v3.3
Copyright (C) 2003-2019 Steve Shaw
Type "help" for a list of commands
The xml is well-formed, applying configuration
hammerdb>
```

The terminal definitely knows how wide it is:
```
sauer@GRODD:~> kubectl -n danny-yugabyte-demo run --rm -i --tty --restart=Never --image webysther/hammerdb hammerdb-danny -- tput cols
80
pod "hammerdb-danny" deleted
```

[...]

PR #306, opened 2022-01-14
Fix getcols can return zero, resulting in divide-by-zero

sm-shaw:
Fix for Issue #289, when run in Kubernetes interactive mode, getcols can return a terminal width of zero. This checks if the columns value is detected, if not it will set a default value of 80.


abondvt89:
Merging Pull Request as voted on by TPC-OSS subcommittee on January 18, 2022.

PR #331, opened 2022-04-05
Add SSL connection options to MariaDB

sm-shaw:
Pull Request enables SSL network connectivity for MariaDB. .pem files should already be created and available in a readable directory and the MariaDB server enabled for SSL. 
Options allow one-way or two-way SSL to be selected. 

sslopts

If SSL is successfully enabled the SSL cipher is reported. Where SSL is not enabled an empty string {} is reported. 

sslrun

Where SSL filenames are given the names are concatenated to the CApath and passed directly for each individual file. If all filenames are empty then the CApath option will be passed instead with the directory name. 
The default SSL Cipher is listed as "server" meaning use the server default. This value can be changed to another available cipher. 

Note: the SSL Cipher requires an update to the mariatcl library and will only work from HammerDB v4.5 unless built from source. 

The same SSL options also available in the CLI. 
```
dbset db maria
diset connection maria_host 127.0.0.1
diset connection maria_port 3306
diset connection maria_ssl true
diset connection maria_ssl_linux_capath /opt/mariadb-10.8.1-linux-x86_64/ssl
diset tpcc maria_driver timed
diset tpcc maria_rampup 2
diset tpcc maria_duration 5
vuset logtotemp 1
loadscript
vuset vu 4
vucreate
vurun
runtimer 600
vudestroy
```

### DIFF
--- a/config/connectpool/mariacpool.xml
+++ b/config/connectpool/mariacpool.xml
@@ -4,6 +4,7 @@
 	<maria_host>host1</maria_host>
         <maria_port>3306</maria_port>
         <maria_socket>/tmp/mariadb.sock</maria_socket>
+	<maria_ssl_options>-ssl false</maria_ssl_options>
 	<maria_user>root</maria_user>
         <maria_pass>maria</maria_pass>
         <maria_dbase>tpcc</maria_dbase>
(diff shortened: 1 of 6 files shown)


abondvt89:

Issue #320, opened 2022-03-04
Running ./hammerdbcli auto <tcl-file> in AWS Lambda function


sm-shaw:
### Discussed in https://github.com/TPC-Council/HammerDB/discussions/319

<div type='discussions-op-text'>

<sup>Originally posted by **joshuajin** March  3, 2022</sup>
I was trying to generate a HammerDB workload from a AWS Lambda function. Specifically running a tcl file to build a schema on a SQL Server database. I tested the ./hammerdbcli auto tcl-file.tcl in a Docker container, that works fine. However after building a AWL Lambda layer with a custom runtime (boostrap), the Lambda function executed ./hammerdbcli -help successfully, but got following error. Has anyone seen these before?

{ [49 bytes data]
* Connection [#0](https://amzn-aws.slack.com/archives/C018BNG2TTK) to host 127.0.0.1 left intact
While loading module "tclreadline"...
no such variable
(read trace on "::env(HOME)")
invoked from within
"variable HISTFILE $::env(HOME)/.tclline_history"
(in namespace eval "::TclReadLine" script line 34)
invoked from within
"namespace eval TclReadLine {
namespace export interact
# Initialise our own env variables:
variable PROMPT "hammerdb>"
variable COMP..."
(file "./modules/tclreadline-1.2.tm" line 28)
invoked from within
"source -encoding utf-8 ./modules/tclreadline-1.2.tm"
("package ifneeded tclreadline 1.2" script)
invoked from within
"package require $m "
END RequestId: a6af8e66-8c1e-44ad-bf67-819228eb1427
REPORT RequestId: a6af8e66-8c1e-44ad-bf67-819228eb1427 Duration: 900.60 ms Billed Duration: 955 ms Memory Size: 128 MB Max Memory Used: 29 MB Init Duration: 54.05 ms 
RequestId: a6af8e66-8c1e-44ad-bf67-819228eb1427 Error: Runtime exited with error: exit status 1
Runtime.ExitErrorRequest ID</div>

sm-shaw:
As per discussion #319 environment variables are not found when running in AWS Lambda. Provide fix so that checks are made before relying on them.

PR #332, opened 2022-04-05
Move history file tracking to interactive

sm-shaw:
Fix for  running ./hammerdbcli auto <tcl-file> in AWS Lambda function #320 
History files only dependent on HOME environment variable when running in interactive mode with prompt.
Using auto flag bypasses history preventing error in AWS lambda


abondvt89:

PR #333, opened 2022-04-07
set MariaDB autocommit to 1 for Monitor VU

sm-shaw:
Minor fix for bug introduced by modifying MariaDB connect procedure for SSL. By default autocommit is set to 0. Monitor VU needs value set to 1 otherwise querying of NOPM value returns 0. 
PR sets autocommit to 1 for the monitor VU to prestore previous behaviour.


abondvt89:
Merging Pull Request as voted on by TPC-OSS subcommittee on April 12, 2022.

PR #334, opened 2022-04-08
fix issue 309 both mysql/postgresql lost connection

dylan2intel:
When I use HammerDB to run a TPC-C benchmarking on mysql and postgresql database, I suffered the db lost connection issue  both on AWS Cloud and on Prem. With a loop investigating in HammerDB's source code, I addressed the root cause that

1. The TPC-C will run 2 stage, the 1st stage is to build schema, while the 2nd is to really run hammer stressing.
2. The db lost connection always occurs at the 1st stage, looking into the code, HammerDB will multiple thread to load the data, suppose you specify N virtual users to build the schema, HammerDb will use 1+N threads to build the schema, the `1` thread will be responsible for create database/tables...then the N threads will go to load table datas, finally back to `1` thread to create stored procedure(Where the issue occurs), since to load a large warehouse(800) table data will cost a long time(1h),  the db connection the '1' thread hold will be timeout, so the step to create stored procedure will fail and always fails here.  per my understanding HammerDB should have introduce a db connection retry method for each db operation. Here my fix is just an workaround.

BTW, to config any timeout parameters it didn't work fine, but to specify a minor warehouse number, the issue not found. However since it is to run performance test, the warehouse cannot be decreased.

### DIFF
--- a/src/mysql/mysqloltp.tcl
+++ b/src/mysql/mysqloltp.tcl
@@ -955,6 +955,7 @@ proc do_tpcc { host port socket count_ware user password db mysql_storage_engine
                 set num_part 0
             }
             CreateTables $mysql_handler $mysql_storage_engine $num_part
+            CreateStoredProcs $mysql_handler
         } else {
             error $mysqlstatus(message)
             return
@@ -1037,7 +1038,6 @@ proc do_tpcc { host port socket count_ware user password db mysql_storage_engine
         }
     }
     if { $threaded eq "SINGLE-THREADED" || $threaded eq "MULTI-THREADED" && $myposition eq 1 } {
-        CreateStoredProcs $mysql_handler
         GatherStatistics $mysql_handler
         puts "[ string toupper $db ] SCHEMA COMPLETE"
         mysqlclose $mysql_handler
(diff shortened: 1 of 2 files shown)


sm-shaw:
Many thanks for the pull request, Yes, that is what was explained in the discussion forum https://github.com/TPC-Council/HammerDB/discussions/310 this is what  I said on the 18th Jan entry:

_The error here is:_

_mysqlexec/db server: Lost connection to MySQL server during query_

_So the monitor virtual user will connect, create the user and tables (load the item table) and then wait for the other virtual users to finish. Once complete, it will then create the stored procedures. It looks like your monitor virtual user was disconnected whilst waiting._

 So to be absolutely clear, the monitor virtual user connects creates the tables and wait for the schemd build is complete until creating the indexes, stored procedures and gathering statistics. 

There are multiple issues with this pull request and it cannot be accepted in its present form. Firstly gathering statistics has to be done after the data load, therefore to fully complete the schema build the monitor user connection needs to be maintained. Creating the indexes before loading the data adds additional overhead. This also addresses just PostgreSQL and MySQL for TPROC-C when HammerDB has TPROC-H for these as well and TPROC-C for additional databases making the schema build process inconsistent across them. 

Therefore this is a workaround for your particular issue of having a timeout somewhere in your configuration disconnecting the monitor virtual user after 1hr (also a TPROC-C build should not be taking anywhere near this long) but also does not gather statistics at the end. Instead of this workaround you should root cause the issue with the timeout, then HammerDB will not have any problems with the schema build.

sm-shaw:
This Pull Request is a workaround for a user issue and is not recommended for approval.

sm-shaw:
Discussed by TPC-OSS subcommittee on April 12, 2022, closed without accepting as addressing a user rather than HammerDB issue.

PR #340, opened 2022-04-19
Escape username and database name in PostgreSQL procedures

aaaronan:
Valid database names and usernames might have upper case characters. However, if
they are passed without proper escaping PostgreSQL will end up referencing the
name using wrong case and later the procedure will error on not finding the
correct username or database name.

Fix this by using double quotes when passing usernames or database names.

All new code of the whole pull request, including one or several files that are
either new files or modified ones, are contributed under the BSD-new license. I
am contributing on behalf of my employer Amazon Web Services, Inc.


sm-shaw:
Merging Pull Request as voted on by TPC-OSS subcommittee on April 20, 2022.

PR #362, opened 2022-05-24
Dependency build automation for Windows

sm-shaw:
This pull request adds Windows build automation to the existing x64 Linux build automation in #323
This enables the compiling of all HammerDB dependencies from a single command.

Prerequisites are an installation of Microsoft Visual Studio 2022 (gcc is also downloaded at build time) and installations of client or server of Db2, MariaDB, MySQL and PostgreSQL including a development environment (headers and libraries).  Environment variables should be set to the location of the database installations as follows:

```
set MARIADB_CONFIG=C:\Program Files\MariaDB\MariaDB Connector C 64-bit
set MYSQL_CONFIG=C:\Program Files\MySQL\MySQL Server 8.0
set PG_CONFIG=C:\Program Files\PostgreSQL\pgsql\bin
set IBM_DB_DIR=C:\Program Files\IBM\SQLLIB
```

Note that the mysql_config command does not exist on Windows and therefore the MYSQL_CONFIG environment variable should be set to the top level MySQL installation directory. All of the specified database installations are required to fully build HammerDB and if one is missing the build will fail. 

Once the environment is set clone or download the HammerDB repository and from the Build\Bawt-2.1.0 directory run the following command:

`Build-Windows.bat x64 vs2022+gcc Setup\HammerDB-Windows.bawt update`

Only the x64 architecture is supported. 

This will download the requisite files and build HammerDB using Visual Studio for Tcl and database libraries and gcc for non-performance critical packages. (Note that Tcl/Tk is also compiled with gcc to provide the configuration environment for the other packages built with gcc but is not installed).  When complete the Bawt Viewer will show the build completion as follows

bawtwindows

In the BawtBuild directory the HammerDB distribution will be built and a zipfile created in the vs2022\x64\Release\Distribution directory.

Note that most of the code changes to enable this pull request are in the database libraries themselves within a new win directory, some source modification to enable the windows build where required as well as the .bawt files providing build instructions for the new files.  As per the discussion in #323 this updated source and .bawt files are not included in the HammerDB repository and are downloaded at build time. 

For this reason the bulk of the code changes required to enable the Windows build automation are not included within this pull request and must be browsed manually after running a build.

### DIFF
--- a/Build/Bawt-2.1.0/Bawt.tcl
+++ b/Build/Bawt-2.1.0/Bawt.tcl
@@ -383,6 +383,20 @@ namespace eval BawtZip {
     # Windows needs 7-Zip and the MSys/MinGW package.
     # Darwin needs 7-Zip.
     proc Bootstrap {} {
+
+    proc Downloadgcc { fileName } {
+        set curlProg [GetCurlProg]
+        set sourceFileUrl [format "%s/%s/%s" [GetBawtUrl] Bootstrap-Windows $fileName]
+        set targetFile [file join [GetBootstrapDir] $fileName]
+        Log "Source file: $sourceFileUrl" 4 false
+        Log "Target file: $targetFile"    4 false
+        set cmd "$curlProg -L $sourceFileUrl -o $targetFile"
+        set result [eval exec "[auto_execok $curlProg ] -L -s $sourceFileUrl -o $targetFile" ]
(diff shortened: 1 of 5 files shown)


abondvt89:
Merging Pull Request as voted on by TPC-OSS subcommittee on May 24, 2022.

PR #366, opened 2022-05-30
Fix RAISEEROR error in Asynch connect

sm-shaw:
Fix for Issue #365 - Error not reported correctly as RAISEERROR variable not found.
Remove RAISEERROR variable for Asynch clients.


sm-shaw:
Pull Request was approved at TPC-OSS subcommittee meeting 31st May 2022.

Issue #371, opened 2022-06-09
SQL Server Schema/Stored Procedures "Optimizing the database"


sm-shaw:
In this [White Paper ](https://www.principledtechnologies.com/Amazon/EC2-R5b-OLTP-competitive-science-0321-v2.pdf) from Principled Technologies running a HammerDB workload, there is the following section:

Optimizing the database:
Amazon Web Services (AWS) provided us with a script that modified the base TPC-C-like database. The script altered some row and page locking settings and modified some stored procedures to improve OPM results for any solution running the benchmark.

As the workload has been modified by AWS, the changes in this section should be evaluated and considered for adoption so that the workload being run by default is consistent.

sm-shaw:
SQL Server Stored Procedures have been implemented as per Principled Technologies/AWS white paper with some minor modifications where profiling showed the changes impacted performance. 
Testing has been done on desktop and laptop only with both in-memory and default versions of the Stored Procedures with the results below including profiling that does have some performance impact. 
Initial testing suggests that the changes improve performance. 

Proposed changes are now open to comments/suggestions/modifications.

SQL Server Developers Edition 15.0.2080.9
Haswell i7 Desktop

CURRENT v4.4 SQL SERVER STORED PROCEDURES/SCHEMA
```
In-memory
Vuser 1:4 Active Virtual Users configured
Vuser 1:TEST RESULT : System achieved 110776 NOPM from 259405 SQL Server TPM
Vuser 1:Gathering timing data from Active Virtual Users...
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
>>>>> SUMMARY OF 4 ACTIVE VIRTUAL USERS : MEDIAN ELAPSED TIME : 239323ms
>>>>> PROC: NEWORD
CALLS: 434379	MIN: 0.249ms	AVG: 0.862ms	MAX: 138.281ms	TOTAL: 374353.281ms
P99: 3.135ms	P95: 1.359ms	P50: 0.740ms	SD: 10416.603	RATIO: 39.106%
>>>>> PROC: PAYMENT
CALLS: 435082	MIN: 0.262ms	AVG: 0.768ms	MAX: 138.132ms	TOTAL: 334212.706ms
P99: 2.869ms	P95: 1.287ms	P50: 0.648ms	SD: 9546.928	RATIO: 34.912%
[...]

PR #372, opened 2022-06-09
SQL Server Schema/Stored Procedure Updates

sm-shaw:
Changes/Updates to SQL Server Schema/Stored procedures as per Issue #371


abondvt89:
Merging Pull Request as voted on by TPC-OSS subcommittee on July 5, 2022.

PR #381, opened 2022-07-08
MSSQLS SET no_ol_i_id to correct value

sm-shaw:
Corrects bug where no_ol_i_id is incorrectly set to a random number from 1,000,000 when it should be 100,000.
Have checked all other database implementations and all are correct apart from this.


memmertoIBM (approved):

sm-shaw:
Accepting Pull Request after approval by TPC-OSS code maintenance team on 9th July 2022

PR #407, opened 2022-08-11
HammerDB: Allow to skip CA for MariaDB

igorkonopko:
This patch adds the ability to pass empty string as the path to certificate authority (CA), when using SSL connection to MariaDB.

Currently when trying to authenticate MariaDB user using SSL key and SSL certificate, HammerDB also requires providing path CA. Providing CA is a recommended solution in production environment, but it is not required. Especially when using self-signed certificates, such a CA cannot be properly verified by OpenSSL and thus it can be rejected, so we want to avoid such a scenario by not specifying path to CA.

### DIFF
--- a/src/mariadb/mariaopt.tcl
+++ b/src/mariadb/mariaopt.tcl
@@ -23,8 +23,11 @@ if { [ file isdirectory $capath ] } {
 if { $maria_ssl_ca eq "" && $maria_ssl_cert eq "" && $maria_ssl_key eq "" } {
 #All of the file entries are blank, use capath only
 } else {
-#CApath is valid, file entries are not blank, always check CA
-if { [ file readable [ file join $capath $maria_ssl_ca ]] } {
+#Two scenarios:
+#-CApath is valid, file entries are not blank, so check CA
+#-CApath is empty when we only want to use cert+key to authenthicate
+#but we don't want to verify CA certificate (useful for self signing scenarios)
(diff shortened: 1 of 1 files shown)


sm-shaw (approved):
Have tested and confirm this is OK to accept. 

currently setting a blank CA will set ssl options as follows:

`set ssl_options { -ssl true  -sslca /opt/mariadb-10.8.1-linux-x86_64/ssl  -sslcert /opt/mariadb-10.8.1-linux-x86_64/ssl/client-cert.pem  -sslkey /opt/mariadb-10.8.1-linux-x86_64/ssl/client-key.pem } ;# Maria SSL/TLS options
`
which gives following error
`Error in Virtual User 1: mariaconnect/db server: SSL connection error: Error while reading file.`

These changes allow the following:
`set ssl_options { -ssl true  -sslcert /opt/mariadb-10.8.1-linux-x86_64/ssl/client-cert.pem  -sslkey /opt/mariadb-10.8.1-linux-x86_64/ssl/client-key.pem } ;# Maria SSL/TLS options`

and also to set CA
`set ssl_options { -ssl true  -sslca /opt/mariadb-10.8.1-linux-x86_64/ssl/ca-cert.pem  -sslcert /opt/mariadb-10.8.1-linux-x86_64/ssl/client-cert.pem  -sslkey /opt/mariadb-10.8.1-linux-x86_64/ssl/client-key.pem } ;# Maria SSL/TLS options`

so the changes make the following allowable in the GUI
mdb

memmertoIBM (approved):

PR #408, opened 2022-08-11
HammerDB: Passwordless MariaDB connection

igorkonopko:
When using SSL connection to MariaDB we are able to authenticate user solely based on SSL subject and issuer. In such a case we need to pass SSL certificate and SSL key to authenticate and we cannot pass user password, because such an user might not have an password set, which is the goal of such an configuration. Currently HammerDB always append '-password' to connection string, so we are not able to connect to MariaDB with user configured as described above. This patch adds the ability to use special keyword "null" in order to skip appending password to connection string. Such a special keyword instead of empty string was chosen in order to avoid breaking other layers of HammerDB and in order to keep the number of changes minimal.

### DIFF
--- a/src/mariadb/mariaolap.tcl
+++ b/src/mariadb/mariaolap.tcl
@@ -66,7 +66,10 @@ proc ConnectToMaria { host port socket ssl_options user password } {
         foreach key [ dict keys $ssl_options ] {
         append connectstring " $key [ dict get $ssl_options $key ] "
         }
-        append connectstring " -user $user -password $password"
+        append connectstring " -user $user"
+        if { [ string tolower $password ] != "null" } {
+        append connectstring " -password $password"
+        }
         set login_command "mariaconnect [ dict get $connectstring ]"
         #eval the login command
         if [catch {set maria_handler [eval $login_command]}] {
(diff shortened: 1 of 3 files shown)


sm-shaw (approved):
Have verified that this works as intended. To work it needs the MariaDB user created without password authentication e.g.
```
MariaDB [(none)]> create user 'steve'@'localhost' require x509;
Query OK, 0 rows affected (0.003 sec)
MariaDB [(none)]> grant all on *.* to 'steve'@'localhost';
Query OK, 0 rows affected (0.003 sec)
```
With connection options and output as follows:
```
-host 127.0.0.1 -port 3306 -ssl true -sslca /opt/mariadb-10.8.1-linux-x86_64/ssl/ca-cert.pem -sslcert /opt/mariadb-10.8.1-linux-x86_64/ssl/client-cert.pem -sslkey /opt/mariadb-10.8.1-linux-x86_64/ssl/client-key.pem -user steve

Vuser 1:Ssl_cipher ECDHE-RSA-AES128-GCM-SHA256
Vuser 1:Processing 1 transactions without output suppressed...
Vuser 1:order status
Vuser 1:1447 ATIONEINGBAR HaQker30A OE -10.00 1239 2022-08-11 16:14:55 3
```
Example options for testing:
mdbopts

Example output:
mdbrun

memmertoIBM (approved):

abondvt89 (approved):

Issue #351, opened 2022-05-06
Add “dropschema” function for GUI, CLI & Webservice


Jiang-Hua:
HammerDB could provide a button/command to drop the database instead of doing it manually, in this way users don’t have to figure out various SQL commands to clean up the environment for different databases.

sm-shaw:
Delete icon in SVG & PNG formats has been added here: https://github.com/TPC-Council/HammerDB/compare/master...sm-shaw:HammerDB:351

PR #430, opened 2022-09-15
Add drop schema functionality

sm-shaw:
Closes Issue #351 

Adds GUI Icon and CLI command to drop the schema created by Build Schema.

For Db2 functionality is added by running external command. To do this correctly we need to add the sqlecrea and the sqledrpd APIs to db2tcl and then we can both create and drop the schema from within HammerDB. 

deleteschema2

### DIFF
--- a/src/db2/db2olap.tcl
+++ b/src/db2/db2olap.tcl
@@ -1195,13 +1195,19 @@ proc ConnectToDb2 { dbname user password } {
 }
 
 proc drop_schema { dbname user password } {
-    set db_handle [ ConnectToDb2 $dbname $user $password ]
-    if { $db_handle ne "" } {
-        set sql "drop database tpch"
-        db2_exec_direct $db_handle $sql
-        db2_disconnect $db_handle
-        puts "TPROC-H schema has been deleted successfully."
+    global tcl_platform
+    if {$tcl_platform(platform) == "windows"} {
+        set cmd "db2cmd -c -w -i db2 drop database tpch"
+    } else {
+        set cmd "db2 drop database tpch"
     }
+
+    if {[ catch {eval exec $cmd} message ]} {
+        error $message
+    } else {
+        puts "TPROC-H Schema has been deleted successfully."
+    }
+
     return
 }
 
--- a/src/db2/db2olap.tcl
+++ b/src/db2/db2olap.tcl
@@ -1156,3 +1156,57 @@ if { $refresh_on } {
     do_tpch $dbname $user $password $scale_factor $RAISEERROR $VERBOSE $degree_of_parallel $total_querysets $myposition
 }}
 }
+
+proc delete_db2tpch {} {
+    global maxvuser suppo ntimes threadscreated _ED
+    upvar #0 dbdict dbdict
+    if {[dict exists $dbdict db2 library ]} {
+        set library [ dict get $dbdict db2 library ]
+    } else { set library "db2tcl" }
+    upvar #0 configdb2 configdb2
+    #set variables to values in dict
+    setlocaltpchvars $configdb2
+    if {[ tk_messageBox -title "Delete Schema" -icon question -message "Ready to delete TPROC-H schema under user [ string toupper $db2_tpch_user ]\n in database [ string toupper $db2_tpch_dbase ]?" -type yesno ] == yes} {
+        set maxvuser 1
+        set suppo 1
+        set ntimes 1
+        ed_edit_clear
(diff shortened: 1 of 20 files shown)


src/db2/db2olap.tcl
```diff
@@ -1156,3 +1156,58 @@ if { $refresh_on } {
...
+        .ed_mainFrame.mainwin.textFrame.left.text fastinsert end {if [catch {package require $library} message] { error "Failed to load $library - $message" }
+if [catch {::tcl::tm::path add modules} ] { error "Failed to find modules directory" }
+if [catch {package require tpchcommon} ] { error "Failed to load tpch common functions" } else { namespace import tpchcommon::* }
+
+proc drop_schema { dbname user password } {
+    global tcl_platform
+        set force "db2 force applications all"
+        set drop "db2 drop database $dbname"
+    if {$tcl_platform(platform) == "windows"} {
+        set prefix "db2cmd -c -w -i"
```
memmertoIBM, line 1192: The use of -c / -w / -i together is redundant.  I believe -i is all you need in this context.

-c = new window, close at terminate
-w = new window, keep open at terminate
-i = use existing window

src/db2/db2oltp.tcl
```diff
@@ -2306,3 +2306,57 @@ switch $myposition {
...
+        .ed_mainFrame.mainwin.textFrame.left.text fastinsert end {if [catch {package require $library} message] { error "Failed to load $library - $message" }
+if [catch {::tcl::tm::path add modules} ] { error "Failed to find modules directory" }
+if [catch {package require tpcccommon} ] { error "Failed to load tpcc common functions" } else { namespace import tpcccommon::* }
+
+proc drop_schema { dbname user password } {
+    global tcl_platform
+        set force "db2 force applications all"
+        set drop "db2 drop database $dbname"
+    if {$tcl_platform(platform) == "windows"} {
+        set prefix "db2cmd -c -w -i"
```
memmertoIBM, line 2342: Same issue here.

src/db2/db2olap.tcl
```diff
@@ -1156,3 +1156,58 @@ if { $refresh_on } {
     do_tpch $dbname $user $password $scale_factor $RAISEERROR $VERBOSE $degree_of_parallel $total_querysets $myposition
 }}
 }
+
+proc delete_db2tpch {} {
```
memmertoIBM, line 1160: Wile I see that you're doing a "db2cmd db2 drop database" here, I could not find where we actually *create* the database, since I would have assumed that there is a similar "db2cmd db2 create database" command.

I do not see db2tcl having any in-built database creation logic either.

What am I missing?

memmertoIBM (changes requested):

src/db2/db2olap.tcl
memmertoIBM: Saw this in the PR:

> To do this correctly we need to add the sqlecrea and the sqledrpd APIs to db2tcl and then we can both create and drop the schema from within HammerDB

But I'm still confused about how we actually create the database today.

src/db2/db2olap.tcl
sm-shaw: This has been tested and updated to just use the -c flag

src/db2/db2oltp.tcl
sm-shaw: This has been tested and updated to just use the -c flag

src/db2/db2olap.tcl
sm-shaw: Right now we ask the user to create the Db2 database manually  - https://www.hammerdb.com/docs/ch04s02.html#d0e1176 and the build just does the load. I did not want to call OS commands as part of the schema build, as this will only work locally. However, as we have the drop schema functionality for all of the other databases this appears to be the best compromise so that it will also work for Db2 in some environments. 

The correct solution is to add the sqlecrea and the sqledrpd APIs to db2tcl and then this could be all done from within HammerDB - however this would take additional time to update the interface. I will create an issue to track this, once sqlecrea and sqledrpd are added to the interface then both the build and drop can be updated to use the correct interface.

memmertoIBM: I created an issue in the db2tcl repo (https://github.com/memmertoIBM/db2tcl/issues/6) to add create/drop functionality.

memmertoIBM (approved):

abondvt89 (approved):

abondvt89:

PR #437, opened 2022-09-23
Improve CLI diset error message

sm-shaw:
Change CLI diset error message so if incorrect dictionary is given, error message reports the dictionary that the key is in or reports that the key is not in any dictionary. 

```
steve@CRANE:/opt/HammerDB-4.5$ ./hammerdbcli
HammerDB CLI v4.5
Copyright (C) 2003-2022 Steve Shaw
Type "help" for a list of commands
hammerdb>dbset db mysql
Database set to MySQL

hammerdb>diset tpcc mysql_ssl true
Dictionary "tpcc" for MySQL exists but key "mysql_ssl" doesn't, key "mysql_ssl" is in the "connection" dictionary
Type "print dict" for valid dictionaries and keys for MySQL

hammerdb>diset connection mysql_ssl true
Changed connection:mysql_ssl from false to true for MySQL

hammerdb>diset connection mysql_sst true
Dictionary "connection" for MySQL exists but key "mysql_sst" doesn't, key "mysql_sst" cannot be found in any MySQL dictionary
Type "print dict" for valid dictionaries and keys for MySQL

hammerdb>
```

### DIFF
--- a/src/generic/gencli.tcl
+++ b/src/generic/gencli.tcl
@@ -517,6 +517,18 @@ proc vuset { args } {
             }
 }}}
 
+proc findakey { key2find dictname } {
+        upvar #0 $dictname $dictname
+        foreach key [dict keys [ set $dictname]] {
+                dict for {k v} [dict get [ set $dictname ] $key] {
+                if { $k eq $key2find } {
+                return $key
+                }
+                }
+        }
+return {}
+}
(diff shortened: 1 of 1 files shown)


abondvt89 (approved):

abondvt89:

Issue #445, opened 2022-10-07
v4.6 keepalive not active when rampup or duration is 0


sm-shaw:
In v4.6 replacement for runtimer the keepalive time is automatically calculated, however if rampup or duration is 0 it reports the error that the values cannot be detected.  The margin for the keepalive time is also fixed but should be user configurable. 
Also in v4.6 automation of loadscript for vucreate only loads the script the first time without doing the creation, it should do both.

PR #446, opened 2022-10-07
Fix for keepalive issue #445

sm-shaw:
Fixes issue #445 where keepalive time is not detected if either are zero and adds user configurable keepalive time.
Also fixes issue with vucreate automatically running loadscript if script is not already loaded.

### DIFF
--- a/src/generic/gencli.tcl
+++ b/src/generic/gencli.tcl
@@ -1049,42 +1049,44 @@ proc buildschema {} {
 }
 
 proc keepalive {} {
-#Routine to keep the main thread alive during vurun
-    global rdbms bm
-    set ka_valid 1
-	if {$bm eq "TPC-C"} {
-#For TPC-C we have a rampup and duration time, find these, check they are valid and call _runtimer automatically with these values
-	 upvar #0 dbdict dbdict
-	 upvar #0 genericdict genericdict
-         if {[dict exists $genericdict commandline keepalive_margin]} {
-                set ka_margin [ dict get $genericdict commandline keepalive_margin]
-                if {![string is entier $ka_margin]} { set ka_margin 10 }
-                } else {
-                set ka_margin 10
-                }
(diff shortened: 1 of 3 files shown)


src/generic/gencli.tcl
```diff
@@ -1051,6 +1055,13 @@ proc keepalive {} {
 	if {$bm eq "TPC-C"} {
 #For TPC-C we have a rampup and duration time, find these, check they are valid and call _runtimer automatically with these values
 	 upvar #0 dbdict dbdict
+	 upvar #0 genericdict genericdict
+         if {[dict exists $genericdict commandline keepalive_margin]} {
+                set ka_margin [ dict get $genericdict commandline keepalive_margin]
+                if {![string is entier $ka_margin]} { set ka_margin 10 }
```
memmertoIBM, line 1061: The if and else branches do the same thing - do we need the condition at all?

memmertoIBM (changes requested):

sm-shaw (commented):
finish review

src/generic/gencli.tcl
memmertoIBM: I understand now.  I had misread the code.

I thought the else on line 1062 was for the if on line 1061, but it's actually for the if on line 1059.  

Can we fix the indentation on lines 1062-1064 so that it is aligned properly with line 1059?

sm-shaw: I have updated the indentation. There is a package included with HammerDB called reformat_tcl. You can do "package require reformat_tcl" and then reformat_src indent filename, and it will update the file with the correct indentation. This should server as a prompt for me to use it to prevent less confusion going forward.

sm-shaw (commented):
requested changes to indentation added.

memmertoIBM (approved):

abondvt89 (approved):

abondvt89:

PR #452, opened 2022-11-03
Add alternative python CLI interface

sm-shaw:
Adds alternative python interface to the CLI for both Linux and Windows, relying on system installed python.
Functionality will allow run scripts to be written in python as well as run interactive sessions. 
The python library will be dependent on a particular version so plan is to use v3.8 on Linux (Unbuntu 20.04), v3.6 on RHEL8 and v3.10 on Windows. However build from source can use any preferred python version instead.
Further discussion in issue #386


sm-shaw:
The background interface to HammerDB from Python is located here https://github.com/sm-shaw/libtclpy so building from source is the best way to test this functionality.

memmertoIBM (approved):

abondvt89 (approved):

abondvt89:

PR #460, opened 2022-11-17
Fixing queries for Columnstore issue 459

### LINKED ISSUE #459: MariaDB TPROC-H columnstore mariaexec/db server: Too many keys specified; max 0 keys allowed
Currently TPOC-H benchmark fails to create benchmark schema on MariaDB when using Columnstore engine. Error as follows:

> ERROR 1069 (42000): Too many keys specified; max 0 keys allowed

Columnstore engine does not allow the usage of primary keys, foreign keys, indexes and BINARY datatype.

mprostko:
This PR adds the logic into MariaDB's OLAP benchmark scripts to use different queries based on the engine used. Resolves #459

### DIFF
--- a/src/mariadb/mariaolap.tcl
+++ b/src/mariadb/mariaolap.tcl
@@ -105,9 +105,9 @@ proc CreateDatabase { maria_handler db } {
     return
 }
 
-proc CreateTables { maria_handler maria_tpch_storage_engine } {
-    puts "CREATING TPCH TABLES"
-    set sql(1) "CREATE TABLE `ORDERS` (
+proc GenerateTableQueries { maria_tpch_storage_engine } {
+    set queries {}
+    lset queries 0 "CREATE TABLE `ORDERS` (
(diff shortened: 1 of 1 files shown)


memmertoIBM (approved):
The differences in table definition boil down to two things:
* column-store doesn't support column-level character set specification (eg, "binary")
* column-store doesn't support creation of PK/FKs

We could use variable replacement to deal with the charset difference, but there's no easy way to deal with the PK/FK differences, so duplicating the CREATE TABLE statements is the best solution at this point.

sm-shaw (approved):
Have tested on a build of MariaDB 10.6 with th columnstore engine and the data is successfully loaded with this change. Note like all columnstores tested to date loading data through inserts is much (much!) slower than a row store and therefore bulk loading is recommended. Neverthless this change is very useful as it gives us the correct table structure to load into.

abondvt89 (approved):

abondvt89:

PR #471, opened 2022-12-08
Updates to jobs for Docker scripts

sm-shaw:
Adds a jobs joblist command that returns a list of all job ids in text. Useful for scripts querying all jobs to get a list to iterate through. Add a help message.
Fix jobs result and jobs timestamp so it reports correctly result and timestamp for all jobs. 
Update jobs format command so accepts upper and lowercase for both text and JSON
Update jobs result so it does not depend on the current set bm value but queries it for each particular job
Add geomean to jobs result for TPROC-H jobs
fix error message in case Jobid already exists
Add jobid return value for buildschema
Fix jobs timestamp was missing conversion to JSON when JSON format was selected
Add return values to python commands for jobs, vurun and buildschema https://github.com/sm-shaw/libtclpy/commit/00830c3f8da00f1cc020d78ddef4b651b844e74b

### DIFF
--- a/modules/jobs-1.0.tm
+++ b/modules/jobs-1.0.tm
@@ -161,18 +161,20 @@ proc jobs { args } {
     1 {
       set param [ lindex [ split  $args ]  0 ]
       if [ string equal $param "result" ] {
-        set alljobs [ jobs ]
+        set alljobs [ getjob "joblist" ]
         foreach jobres $alljobs {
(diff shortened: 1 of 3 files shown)


memmertoIBM (approved):

abondvt89 (approved):

abondvt89:

Issue #472, opened 2022-12-10
hammerdbcli auto script.tcl - Not Exists


dineshtessell:
Hi Team,

Thanks for the HammerDB benchmark tool, and its great.

As part of exploring the `HammerDB`, we found an issue where the `hammerdbcli` is going to hang.

We tested this functionality, by shutting down the PostgreSQL instance.

Then ran the `hammerdbcli` as like below.

```
$ ./hammerdbcli auto pgschema.tcl
HammerDB CLI v4.6
Copyright (C) 2003-2022 Steve Shaw
Type "help" for a list of commands
Initialized SQLite on-disk database /tmp/hammer.DB using existing tables (430,080 KB)
Database set to PostgreSQL
Benchmark set to TPC-C for PostgreSQL
Value disable for connection:pg_sslmode is the same as existing value disable, no change made
Script cleared
Deleting schema with 1 Virtual User
Do you want to delete the TPCC TPROC-C schema and role TPCC
 in host LOCALHOST:5432 under user POSTGRES?
Enter yes or no: replied yes
Vuser 1 created - WAIT IDLE
Vuser 1:RUNNING
Vuser 1:Connection to database failed connection to server at localhost (127.0.0.1), port 5432 failed: Connection refused Is the server running on that host and accepting TCP/IP connections? connection to server at localhost (::1), port 5432 failed: Cannot assign requested address Is the server running on that host and accepting TCP/IP connections?
Vuser 1:FINISHED FAILED
ALL VIRTUAL USERS COMPLETE
vudestroy success

...
...
...


^C
```

pgschema.tcl as below

`
$ cat pgschema.tcl
dbset db pg
dbset bm TPC-C
diset connection pg_sslmode disable

deleteschema
vudestroy

vuset logtotemp 1
diset tpcc pg_storedprocs true
diset tpcc pg_count_ware 4
diset tpcc pg_num_vu 1
diset tpcc pg_raiseerror true
buildschema
quit
`

Would you please let us know if we miss anything here. If postgresql is up and running, then the schema creation is successful, 
and the script exit status is as 0, but it is not the case when postgresql is shutdown.

Please let me know if i miss anything here.

Thanks

sm-shaw:
Hi, 
[...]

PR #473, opened 2022-12-12
Prevent CLI hang when scripts finishes before enters event loop

sm-shaw:
Fix for Issue #472 

I have used the same clause for consistency even though timevar is not set previously in waittocomplete and is set to 0 in runtimer so the same expression will work if it does not exist or is not set to 1.

`![ info exists timevar ] || $timevar != 1`

Note the || operator is a  "short-circuiting" operator in Tcl. Meaning, they are evaluated left-to-right, and evaluation stops when the expression is known. So if the timevar variable does not exist it will not check $timevar != 1 and then fail because timevar is not set. 

This change fixes the scenario in #472 where a script that is being run fails so quickly it has already set the variable to exit the event loop before the loop has been entered, and therefore once the loop is entered it then waits forever. This would not be encountered under normal circumstances and counts as an edge case, but nevertheless an inconvenience if not knowing why this occurs.  

To fix this we add the check to see whether the timevar variable has been set as the script failed so quickly we didn't have time to enter the loop, if it has, we are in the scenario above and therefore do not set the loop but instead return as there is nothing to wait for.

### DIFF
--- a/src/generic/gencli.tcl
+++ b/src/generic/gencli.tcl
@@ -1467,7 +1467,9 @@ proc _waittocomplete {} {
     }
     set wcomplete "false"
     wait_to_complete_loop
-    vwait timevar
+    if {![ info exists timevar ] || $timevar != 1 } { 
+    	vwait timevar
+    	}
     return
 }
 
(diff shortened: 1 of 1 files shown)


memmertoIBM (approved):

abondvt89 (approved):

abondvt89:

PR #481, opened 2022-12-29
Query update for tpch columnstore benchmark

### LINKED ISSUE #480: MariaDB Columnstore TPROC-H query failures
Running TPROC-H queries on MariaDB's Columnstore result in following errors:

Query 2:
> Internal error: MCS-3012: Scalar filter and semi join are not from the same pair of tables.

Query 17:
> Internal error: MCS-3012: Scalar filter and semi join are not from the same pair of tables.

Query 19:
> Internal error: MCS-1000: 'LINEITEM' and 'PART' are not joined.

mprostko:
Closes #480

This fix updates the queries for MariaDB's OLAP tests and fixes three failing queries. Fixed queries were backported from [here](https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/5923beeab9397aa22563ff7b1f0f31ad8054bae6/mysql-test/columnstore/csinternal/include/dbt3/dbt3_performance_queries_1g.sql).


memmertoIBM (approved):

memmertoIBM:
> In particular, retaining the original queries where we can and only changing them is closer to the original TPC-H specification that the workload is derived from. (As the specification doesn't permit rewriting)

The addition of correlation names (AS CLAUSE) is allowed as a "minor query modification" as per TPC-H 2.2.3.3 h).

That said, since the existing query text is not ambiguous in the first place, the need for additional correlation names in the MariaDB columnstore implementation indicates a problem with the MariaDB columnstore query parser, so this latest approach is best.

abondvt89 (approved):

PR #503, opened 2023-02-02
Fix TPROC-H query 5 for Columnstore

igorkonopko:
MariaDB Columnstore requires one more modification in TPROC-H queries in order to produce proper results. It requires to modify query 5 in order to return proper data instead of empty set.
This patch adds  "+ 0" to query, so that engine will see that the condition as a cross table comparison rather than a join to avoid the circular join error.
Based on: https://github.com/mariadb-corporation/mariadb-columnstore-engine/blob/develop/mysql-test/columnstore/include/dbt3/dbt3_performance_queries_1g.sql


sm-shaw (approved):
I have tested the change on both mariadb columnstore and innodb and they both work as intended.

abondvt89 (approved):

PR #504, opened 2023-02-02
Fix calculation of geomean in TPROC-H

igorkonopko:
Query 15 in TPROC-H is a multi-statement query, which consists of two parts: creating the view and querying that view. Due to that it is handled using separate code then all the other queries. Currently, the execution time of query 15 is not calculated as a part of geometric mean value, which is the metric for that test.

This patch fixes that to include  query execution time in geometric mean metric. Creating/dropping the view is still not taken into that calculation, so no changes here.


sm-shaw (approved):
I have tested the change on both mariadb columnstore and innodb and they both work as intended.

memmertoIBM (approved):

abondvt89 (approved):

Issue #508, opened 2023-02-07
Include the sample scripts as part of HammerDB Distribution package


pooja-jain-17:
Can we have the [scripts](https://github.com/TPC-Council/HammerDB/tree/master/Docker/scripts) directory placed in the HammerDB main directory and be bundled in HammerDB package. Rather than being part of only Docker image. 

This will accomplish the following objectives:
1.	Anyone using HammerDB, with or without Docker will have access to the scripts. 
2.	Anyone can pickup the HammerDB Dockerfile and build their own image without having to additionally download the scripts folder. Dockerfile will be sufficient, enabling a straightforward deployment. 
3.	This will make it easier to create database specific Dockerfiles as outlined in Issue #[506](https://github.com/TPC-Council/HammerDB/issues/506)

sm-shaw:
This is a very good suggestion and one we should implement to include the example scripts in the main directory.

sm-shaw:
Also, if we are including the sample scripts in the main HammerDB directory, then we should also add a Windows powershell script alongside the current bash shell - this way the scripts will be able to run on Windows also. To keep things more straightforward, I will open a new issue to track this.

PR #509, opened 2023-02-07
Adding sample scripts to HammerDB home directory

pooja-jain-17:
Added the sample scripts directory to home folder of HammerDB as described in Issue #508 

Will delete scripts in Docker folder when we update the Dockerfile to version 4.7 in subsequent updates. 
Doing so now will break the auto-builds and docker images. Thanks.

### DIFF
--- a/scripts/python/generic/generic_tprocc_result.py
+++ b/scripts/python/generic/generic_tprocc_result.py
@@ -1,0 +1,68 @@
+redirect_stdout = (r'''proc stdout { switch { file "" } } {
+     if { ! [ llength [ info command __puts ] ] && \
+            [ string equal off $switch ] } {
+        rename puts __puts
+        if { [ string length $file ] } {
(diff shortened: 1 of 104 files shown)


sm-shaw (changes requested):
This looks good. We should also remove them from the Docker folder as well at the same time with another commit to this PR, so we don't have the same scripts in 2 different locations, as they could then diverge over time.

sm-shaw (approved):

sm-shaw:
Please ignore initial review, I re-read 

_Will delete scripts in Docker folder when we update the Dockerfile to version 4.7 in subsequent updates.
Doing so now will break the auto-builds and docker images. Thanks._

So yes, this is fine as is and we can delete the ones in the Docker folder with a future PR

memmertoIBM (approved):
Approving on the strength of Steve's review.

abondvt89 (approved):

PR #513, opened 2023-02-20
Fix fails to delete if SQLite file open for read on Windows closes #512

### LINKED ISSUE #512: HammerDB 4.7 fails to start on Windows if 4.6 SQLite files present
HammerDB checks version number against the SQLite files and deletes and recreates them in case of mismatch. To check  the version it opens the generic.db file. On Linux we can delete a file open for read, on Windows we get permission denied, this is not caught on startup and stops HammerDB starting on Windows. 

The GUI, CLI and WS init files need updating to close SQLite before deleting the files and also to catch any error messages when trying to delete. 

Example as follows:  
```
#SQLite found, check whether the schema versions from SQLite and XML are consistent
if { $sqlite_hdb_version ne $hdb_version } {
puts "The existing SQLite DBs are from version $sqlite_hdb_version. SQLite DBs will be reset to $hdb_version."
# Close SQLite before deleting else get permission denied on Windows
if { [catch {hdb close} message]} { puts "Failed to close SQLite file: $message" }
foreach { dbname } { generic database db2 mariadb mssqlserver mysql oracle postgresql } {
set dbfile [ CheckSQLiteDB $dbname ]
if { [catch {file delete $dbfile} message]} { puts "Error deleting SQLite file from $sqlite_hdb_version: $message" }
}

```

sm-shaw:
Closes issue #512 and enables v4.7 to start if v4.6 SQLite files already exist.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1046,6 +1046,7 @@ Version 4.7 Feb 2023
 
 Pull Requests & Issues
 
+HammerDB 4.7 fails to start on Windows if 4.6 SQLite files present #513 (#512)
 Fix calculation of geomean in TPROC-H #504
 Fix TPROC-H query 5 for Columnstore #503
 Use db2tcl to create and drop database for DB2 #501 (#431)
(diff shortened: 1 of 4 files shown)


memmertoIBM (approved):

abondvt89 (approved):

abondvt89:

PR #531, opened 2023-04-05
Repair ora_compatible OSTAT procedure

mw2q:
An extra EXCEPTION clause was inserted at some point.  This syntax error caused the creation of the database to also miss the creation of the SLEV stored procedure when using EPAS in Oracle compatibility mode with stored procedures.

Remove the extra condition "WHEN NO_DATA_FOUND".

It appears to have be broken since before the git repository was created.

### DIFF
--- a/src/postgresql/pgoltp.tcl
+++ b/src/postgresql/pgoltp.tcl
@@ -415,9 +415,10 @@ proc CreateStoredProcs { lda ora_compatible citus_compatible pg_storedprocs } {
             ORDER BY o_id DESC)
             WHERE ROWNUM = 1;
             EXCEPTION
-            WHEN NO_DATA_FOUND THEN
+            WHEN serialization_failure OR deadlock_detected OR no_data_found THEN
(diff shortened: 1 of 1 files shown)


sm-shaw (approved):
Thank-you. The oracle compatible PostgreSQL stored procedures are fairly old, and both the PostgreSQL and Oracle ones have been updated since. But if they continue to work to test the Oracle compatibility functionality, then it is fine to still leave it there, although I would not expect it to perform as well as the non-Oracle compatible ones.

mw2q:
Sounds good.  I was going to make a couple passes.  A quick fix so it'll work, and then updating the procedures to match the Oracle code better.  I think we want to match the Oracle code as opposed to the PostgreSQL code because I think showing Oracle compatibility is more interesting.

mw2q:
Oops, I pushed a bad branch.  Amended.

jamiere-msft (approved):

abondvt89 (approved):

Issue #534, opened 2023-04-14
Configuring Schema Build Options doc bug


sm-shaw:
In the following section of the documentation https://www.hammerdb.com/docs/ch04s03.html

The row in the options table for "Virtual Users to Build Schema" says:

The  Virtual Users to Build Schema is the number of Virtual Users to be  created on the Load Generation Server that will complete your  multi-threaded schema build. You should set this value to either the  number of warehouses you are going to create (**You cannot set the number  of virtual users lower than the number of warehouses value**) or the  number of cores/Hyper-Threads on your Load Generation Server. If you  have a significantly larger core/Hyper-Thread count on your Database  Server then also installing HammerDB locally on this server as well to  run the schema build can take advantage of the higher core count to run  the build more quickly.

The highlighted section is a typo error and should say instead:

**You cannot set the number of virtual users _higher_ than the number of warehouses value**

In other words, you can build 10 warehouses with 10 VUs and you can build 100 warehouses with 10 VUs but you cannot build 10 warehouses with 20 VUs.  This is because each warehouse can be built by 1 VU only. the GUI will automatically restrict the options if the user tries to increase the number of VUs higher than the warehouse count e.g.

![image](https://user-images.githubusercontent.com/38044085/232007356-421811af-d600-4708-93d7-6b436921a4c3.png)

sm-shaw:
As a straightforward typo this change has already been made in the online documentation here https://www.hammerdb.com/docs/ch04s03.html

jamiere-msft:
Looks good.

PR #535, opened 2023-04-14
Fix doc bug for VU/warehouse count

sm-shaw:
Fix for issue #534 doc advises VU count can't be lower than warehouse count when it should say higher.


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

Issue #23, opened 2018-12-18
Auto scale factor - TPC-H


abondvt89:
Auto set of scale factor in driver to reduce errors (based on supplier 10K row count)

PR #539, opened 2023-04-17
Auto scale factor for TPC-H (issue 23)

Jiang-Hua:
Auto scale factor for TPC-H (issue 23)


sm-shaw (approved):
Have tested the changes and works correctly to resolve issue #23

jamiere-msft (approved):

abondvt89 (approved):

Issue #538, opened 2023-04-16
Arithmetic overflow for larger databases


Alex-Zarenin:
**Describe the bug**
A script generated for SQL Server TPPROC-C includes the following statements:

if {[catch {set rows [ odbc allrows "**select sum(d_next_o_id) from district**" ]} message ]} {
                error "Failed to query district table : $message"

For large size test databases (30,000 warehouses, 75,000 warehouses) this statement results in arithmetic overflow. To address this issue, I have to manually modify the script before each run to replace statement above with:

if {[catch {set rows [ odbc allrows "**select sum(cast(d_next_o_id as _bigint_)) from district**" ]} message ]} {
                error "Failed to query district table : $message"

Casting **d_next_o_id** to **_bigint_** addressing the issue with arithmetic overflow. It would be helpful to include this cast into the script generator so that users do not have to do this manual change. BTW, this statement appears **twice** in the script.

**To Reproduce**
1. Generate large test database - 30,000 warehouses or more
2. Run performance test with 500 virtual users using Autopilot
3. See error

**Expected behavior**

I would expect that test run successfully independent of the size of the test database.

**HammerDB Version (please complete the following information):**
 - Version: 4.6
 - Build: 

**HammerDB Interface (please complete the following information):**
 - UI: GUI

**Operating System (please complete the following information):**
 - Server OS: Windows Server 2019
 - Client OS: Windows Server 2019

**Database Server (please complete the following information):**
 - Database name: SQL Server 2019

**Database Client (please complete the following information):**
 - Database client name: [ODBC Driver 17 for SQL Server

**Additional context**

Alex-Zarenin:
I just created a pull request for this issue: https://github.com/TPC-Council/HammerDB/pull/541

PR #540, opened 2023-04-17
Update mssqlsoltp.tcl

Alex-Zarenin:
Replaced  "select sum(d_next_o_id) from district" with "select sum(cast(d_next_o_id as bigint)) from district" in 4 places in the file to avoid arithmetic overflow when running HammerSDB against test database with large number of warehouses.


Alex-Zarenin:
This pull request addresses the issue https://github.com/TPC-Council/HammerDB/issues/538

PR #541, opened 2023-04-17
Update mssqlsoltp.tcl

Alex-Zarenin:
Replaced  "select sum(d_next_o_id) from district" with "select sum(cast(d_next_o_id as bigint)) from district" in 4 places in the file to avoid arithmetic overflow when running HammerSDB against test database with large number of warehouses.

This pull request addresses the issue https://github.com/TPC-Council/HammerDB/issues/538


sm-shaw (approved):
I have built from source and run tests and everything works as expected.

jamiere-msft (approved):

abondvt89 (approved):

Issue #561, opened 2023-06-15
Build schema fails for password with semicolon - command injection


marjus45:
**Describe the bug**

The test run fails to build the schema when providing password with semicolon (`;`), e.g. `abc;xyz`, with the following error message:

```
Changed tpcc:pg_superuserpass from postgres to abc for PostgreSQL
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<string>", line 17, in <module>
  File "/home/hammerdb/HammerDB-4.7/lib/tclpy0.4/hammerdb.py", line 51, in diset
    eval_hammerdb_command('diset',*args)
  File "/home/hammerdb/HammerDB-4.7/lib/tclpy0.4/hammerdb.py", line 12, in eval_hammerdb_command
    output = tclpy.eval(to_hdb)
RuntimeError: invalid command name "xyz"
```

**To Reproduce**

1. Start a PostgreSQL database and set a user with password `abc;xyz`
1. Run the docker image from tpcorg/hammerdb
    ```
    docker run -it tpcorg/hammerdb bash
    ```
1. Edit the `scripts/python/postgres/tprocc/pg_tprocc_buildschema.py` and set the `pg_superuserpass` as `abc;xyz`
    ```
    ...
    diset('tpcc','pg_superuserpass','abc;xyz')
    ...
    ```
1. Edit any other configuration settings to point to your PostgreSQL database and run the buildschema script
    ```
    ./hammerdbcli py auto scripts/python/postgres/tprocc/pg_tprocc_buildschema.py
    ```

**Expected behavior**

I expect for the schema to be build successfully.

**HammerDB Version (please complete the following information):**
 - Version: 4.7
 - Build: Release download (digest:sha256:d0df547d32a028a884adc3d7eb39b28e8a074dd42811eea61b4404c186307c63)

**HammerDB Interface (please complete the following information):**
 - CLI

**Operating System (please complete the following information):**
 - Client OS: Container which uses Ubuntu 20.04

**Database Server (please complete the following information):**
 - Database name: PostgreSQL
 - Database Release Version: 15.3

**Additional context**

I narrowed down a bit the problem:
1. Enter `hammerdbcli` python environment
    ```console
    root@0db65de51c49:/home/hammerdb/HammerDB-4.7# ./hammerdbcli python
    ```
1. Set PostgreSQL variables:
    ```console
    hammerdb>>>dbset('db','pg')
    Database set to PostgreSQL
    
[...]

PR #566, opened 2023-06-19
Add password blank to GUI

sm-shaw:
As partly prompted by discussion in  #561 on other area adds password blanking for GUI fields.
May seem superfluous as the passwords will still be entered in the script editor in clear text, stored in the config dict and the SQLite database which is why it hasn't been done before.  
However may be of benefit when sharing screen shots of the GUI and don't want to share password field.


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

Issue #577, opened 2023-07-01
Numeric field overflow in warehouse


YongChoEDB:
**Describe the bug**
After running the tpcc workload for several hours, the Postgres log starts printing `ERROR: numeric field overflow`.

**To Reproduce**
1. Run the following script until completion.
```
 dbset db pg
 
 diset connection pg_host xxx
 diset tpcc pg_superuserpass xxx
 diset tpcc pg_user tpcc
 diset tpcc pg_pass tpcc
 diset tpcc pg_dbase xxx
 
 diset tpcc pg_count_ware 1
 diset tpcc pg_num_vu 1
 
 buildschema
 waittocomplete
```

2. Run the following script against the same database. Let it run for several hours.
```
 dbset db pg
 
 diset connection pg_host xxx
 diset tpcc pg_superuserpass xxx
 diset tpcc pg_user tpcc
 diset tpcc pg_pass tpcc
 diset tpcc pg_dbase xxx
 
 diset tpcc pg_count_ware 1
 diset tpcc pg_num_vu 1
 diset tpcc pg_total_iterations 100000000
 
 loadscript
 
 
 vuset vu 1
 vuset delay 10
 vuset repeat 10
 vuset showoutput 0
 
 vurun
 waittocomplete
```

3. Observe the Postgres log for the following messages.

```
2023-07-01 20:04:23.928 UTC [4403] STATEMENT:  prepare payment (INTEGER, INTEGER, INTEGER, INTEGER, INTEGER, INTEGER, NUMERIC, VARCHAR) AS select payment($1,$2,$3,$4,$5,$6,$7,'$8','0',0)
2023-07-01 20:04:23.944 UTC [4403] ERROR:  numeric field overflow
2023-07-01 20:04:23.944 UTC [4403] DETAIL:  A field with precision 12, scale 2 must round to an absolute value less than 10^10.
2023-07-01 20:04:23.944 UTC [4403] CONTEXT:  SQL statement "UPDATE warehouse
                   SET w_ytd = w_ytd + p_h_amount
                 WHERE w_id = p_w_id
                 RETURNING w_street_1, w_street_2, w_city, w_state, w_zip, w_name"
        PL/pgSQL function payment(integer,integer,integer,integer,integer,integer,numeric,character varying,character varying,numeric) line 60 at SQL statement
[...]

PR #585, opened 2023-08-21
Update precision of w_ytd to prevent numeric field overflow

sm-shaw:
As discussed in Issue #577 update to prevent numeric field overflow with long-running benchmarks. Changed precision of W_YTD from 12 to 16 in WAREHOUSE table. (12 is the default in the TPC specification).


jamiere-msft (approved):

sm-shaw:
As an additional check on the numeric overflow I checked w_ytd start value, and it has an additional 0 so should be 300000.00, and instead it is incorrectly set to 3000000.00. This is not expected to impact performance and initial PostgreSQL precision change should remain as field overflow far exceeds even the higher value. However, commit https://github.com/TPC-Council/HammerDB/pull/585/commits/b7015cdbcb64b84aeee3c61147a18fdc48710594 brings w_ytd back in line with sum of d_ytd for a warehouse which it should be.

abondvt89 (approved):

abondvt89:

PR #592, opened 2023-09-08
Added use_bcp default as true and enabled bcp with windows authentication

krithikasatish:
The use_bcp option was by default set to false because bcp did not work through windows authentication. Bcp statement is now modified based on authentication type, with the -T flag used in place of a specified username and password when the connection is through windows authentication. This ensures that the bcp command will use the Windows logged in user's credentials.

### DIFF
--- a/config/mssqlserver.xml
+++ b/config/mssqlserver.xml
@@ -23,7 +23,7 @@
             <mssqls_imdb>false</mssqls_imdb>
             <mssqls_bucket>1</mssqls_bucket>
             <mssqls_durability>SCHEMA_AND_DATA</mssqls_durability>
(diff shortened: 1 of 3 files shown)


jamiere-msft (approved):

sm-shaw (approved):
On testing, virtual users would error with following message:
`Error in Virtual User 2: Error: can't read "authentication": no such variable`
This is because it checks the value of the authentication variable but this variable is not set in this proc. Added upvar so that the authentication variable setting can be queried without error. 
With fix test shows that bcp load works with Windows authentication meaning the option can be the default as it is faster.

sm-shaw:
Tested on all combinations of Windows and SQL authentication with and without BCP selected

abondvt89 (approved):

PR #556, opened 2023-06-08
MySQL/MariaDB L_ORDERKEY to BIGINT

sm-shaw:
Fix for Issue #551 and discussion #550

PR #599, opened 2023-09-15
Fix MySQL/MariaDB TPROC-H build FK error

sm-shaw:
Bug was introduced into MySQL/MariaDB TPROC-H  schema with build https://github.com/TPC-Council/HammerDB/pull/556 resulting in following error creating tables
```
Error in Virtual User 1: Error: mysqlexec/db server: Referencing column 'L_ORDERKEY' and referenced column 'O_ORDERKEY' in foreign key constraint 'LINEITEM_ibfk_1' are incompatible.
Error in Virtual User 1: Error: mariaexec/db server: Can't create table `tpch`.`LINEITEM` (errno: 150 "Foreign key constraint is incorrectly formed")
```
PR changes O_ORDERKEY to bigint  O_ORDERKEY BIGINT NOT NULL so FK is correctly formed. Fixed in both MySQL and MariaDB.


abondvt89 (approved):

jamiere-msft (approved):

abondvt89:

Issue #597, opened 2023-09-14
HammerDBcli does not work correctly when added to PATH and called from other directory


akantak:
**Is your feature request related to a problem? Please describe.**
When adding HammerDB home to the PATH (on Linux), and calling hammerdbcli from a different directory, it throws an error, as it is using relative paths, like `./bin/tclsh8.6`
```
$ echo $PATH
(...):/home/akantak/HammerDB-4.8
$ pwd
/home/akantak
$ hammerdbcli
/home/akantak/HammerDB-4.8/hammerdbcli: 12: exec: ./bin/tclsh8.6: not found
$
```

**1. Describe the solution you'd like**
Add code to change the directory to HammerDB "home", in `hammerdb`, `hammerdbcli` and `hammerdbws`, right at the beginning of the scripts, in line 3. Example code:
```
## \
cd "$(dirname "$0")"
```

**2. Describe alternatives you've considered**
The other approach here can be to define in the same line 3 a variable called HAMMERDB_HOME, and change all relative paths to full paths, like `${HAMMERDB_HOME}/bin/tclsh8.6` instead of `./bin/tclsh8.6`, but it would require more changes in the code.
**3. Describe alternatives you've considered**
A third approach could be to suggest adding to the path both `${HAMMERDB_HOME}` and `${HAMMERDB_HOME}/bin` and get rid of relative paths, using `tclsh8.6` instead of `./bin/tclsh8.6`. But it can stop working when somebody has not added Hammerdb to PATH.

**Additional context**
I can submit a PR on my own, but wanted to discuss a solution first.

akantak:
I submitted a draft, but I have still one doubt.
in agent/agent line 10, it can be either `${0##*/}` so extracting file name only from the path, or hardcoded `./agent` like it is in agent/agent.bat. Because when we call, ex. ./agent/agent, but change dir, this path is no longer valid, and should be changed to ./agent.
Also, not sure if anybody would start an agent per se, with an added ${HAMMERDB}/agent to the path too. Is it a valid use case @sm-shaw ?

Until we figure this out, I marked the PR as draft.

akantak:
And the same applies to all calls when using `${0}`, like in hammerdbcli, or hammerdb. It should be either changed to `${0##*/}` or hardcoded called file name.
These changes are not considered in the PR draft yet.

sm-shaw:
[...]

PR #601, opened 2023-09-21
Fix calling HammerDB added to PATH

akantak:
Fixing #597

### DIFF
--- a/agent/agent
+++ b/agent/agent
@@ -7,7 +7,7 @@ export LD_LIBRARY_PATH=.././lib:.././lib64:$LD_LIBRARY_PATH
 # \
 export PATH=.././bin:$PATH
 # \
-exec .././bin/tclsh8.6 $0 ${1+"$@"}
+exec .././bin/tclsh8.6 ${0##*/} ${1+"$@"}
 # \
 exit
 ########################################################################
--- a/agent/agent
+++ b/agent/agent
@@ -1,6 +1,8 @@
 #!/bin/sh
 ########################################################################
 # \
+cd "$(dirname "$0")"
+# \
 export LD_LIBRARY_PATH=.././lib:.././lib64:$LD_LIBRARY_PATH
 # \
 export PATH=.././bin:$PATH
--- a/agent/agent.bat
+++ b/agent/agent.bat
@@ -1,3 +1,4 @@
 @echo off
+cd /D "%~dp0"
 set path=..\.\bin;%PATH%
 START tclsh86t .\agent
--- a/agent/mpstat.bat
+++ b/agent/mpstat.bat
@@ -1,5 +1,6 @@
 ::if 0 {--*-tcl-*--
  @echo off
+cd /D "%~dp0"
 set path=..\.\bin;%PATH%
  if "%OS%" == "Windows_NT" goto WinNT
  tclsh86t "%0" %1 %2 %3 %4 %5 %6 %7 %8 %9
--- a/hammerdb
+++ b/hammerdb
@@ -1,6 +1,8 @@
 #!/bin/sh
 ########################################################################
 # \
+cd "$(dirname "$0")"
+# \
 export LD_LIBRARY_PATH="./lib:$LD_LIBRARY_PATH"
 # \
 export PATH="./bin:$PATH"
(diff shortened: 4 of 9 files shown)


akantak:
discussion happen in #597

sm-shaw (changes requested):
This looks like its moving in the right direction but needs a bit more debugging and testing. If I try and run hammerdb from another directory, I get: 
`
Error in startup script: couldn't read file "./HammerDB-4.8-600/hammerdb": no such file or directory
`
This is because it is trying to execute the absolute path rather than the relative one. 
`exec wish8.6 -file $0 ${1+"$@"}`

So this looks like it needs the same change as the agent e.g.
`exec .././bin/tclsh8.6 ${0##*/} ${1+"$@"}`

However, there are other tests I did that get broken with these changes such as running the automated scripts such as:
```
./scripts/tcl/maria/tprocc/maria_tprocc.sh
BUILD HAMMERDB SCHEMA
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-
can't read "LD_LIBRARY_PATH": no such variable
    while executing
"export LD_LIBRARY_PATH="./lib:$LD_LIBRARY_PATH""
    (file "hammerdbcli" line 6)
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-
RUN HAMMERDB TEST
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-
can't read "LD_LIBRARY_PATH": no such variable
    while executing
"export LD_LIBRARY_PATH="./lib:$LD_LIBRARY_PATH""
    (file "hammerdbcli" line 6)
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-
DROP HAMMERDB SCHEMA
can't read "LD_LIBRARY_PATH": no such variable
    while executing
"export LD_LIBRARY_PATH="./lib:$LD_LIBRARY_PATH""
    (file "hammerdbcli" line 6)
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-
HAMMERDB RESULT
can't read "LD_LIBRARY_PATH": no such variable
    while executing
"export LD_LIBRARY_PATH="./lib:$LD_LIBRARY_PATH""
    (file "hammerdbcli" line 6)
```
So it needs testing with the hammerdbcli auto argument with both Tcl and Python options. 
There is also an auto argument to the GUI that will load a file and run the autopilot that needs testing as well. 

It needs testing from local and remote as another error I saw running locally was the following: 
```
./hammerdb
While loading module "tkpath"...
can't find package tkpath
    while executing
"package require $m "
```
On Linux a quick hack showed something like this worked a bit better, however I cannot guarantee it satisifes all test conditions:
```
#!/bin/sh
########################################################################
# \
cd "$(dirname "$0")"
# \
export LD_LIBRARY_PATH="${0##*/}/lib:./lib:$LD_LIBRARY_PATH"
# \
export PATH="${0##*/}/bin:./bin:$PATH"
# \
exec wish8.6 -file ${0##*/} ${1+"$@"}
# \
exit
########################################################################
```
On Windows a basic test worked, however I have not tested all scenarios such as running the scripts from Powershell. 

So if this has been thoroughly tested then the idea is good, however we need to be absolutely sure that it doesn't break some of the functionality inadvertently as this is the reason it has not been done already as changing to the home directory before running we know for certain every test is passed, so we need to have the same confidence with these changes.

akantak:
@sm-shaw Yes, I'm aware of it, that's why it is marked as a draft, to present the flaws in that approach. The relative paths are an issue when using `cd` at the beginning of a script. That's why I continued a discussion at https://github.com/TPC-Council/HammerDB/issues/597 on which option to choose from the one that I shared. I will be back on 02.10 to continue the discussion.

akantak:
will continue discussion here:
https://github.com/TPC-Council/HammerDB/issues/597
for now, due to lack of my time, closing this PR as there were flaws discovered in presented solution.

PR #602, opened 2023-09-21
Add MariaDB/MySQL No Stored Procedures option to TPROC-C

sm-shaw:
PR adds feature for MariaDB/MySQL allowing a "No Stored Procedures" option which if selected means that the stored procedures are not called and the application logic runs in the client making a separate call from client to server for each SQL statement. The goal has been to replicate as closely as possible the SQL in the existing MariaDB/MySQL stored procedures which are the same in a client form. 
This allows comparisons to be done between client SQL and stored procedures and also benefits scenarios looking to measure increased network traffic. 
Tests showed at maximum 30% and 55% higher performance for stored procedures over client SQL for MariaDB and MySQL respectively testing over a TCP connection on the local host (instead of using sockets) and all tests completed successfully.


sm-shaw:
Note for review that when chosen feature does an inline edit of the existing editor to replace the stored procedures with the client SQL rather than duplicating multiple copies of the same code using the lines below in the commit to delete the existing section and replace it with the client version built depending on the options chosen such as test/time sync/async. 
```
#Delete stored procedures
.ed_mainFrame.mainwin.textFrame.left.text fastdelete $index_sp_1 $index_sp_2+1l
#Insert no stored procedures version
.ed_mainFrame.mainwin.textFrame.left.text fastinsert $index_sp_1 "$neword_no_sp \n\n $pay_no_sp \n\n $ostat_no_sp \n\n $deliv_no_sp \n\n $stock_no_sp \n\n"
```
So the procs will not look complete in the commit only when the option is chosen in the editor window.

abondvt89 (approved):

abondvt89:

PR #613, opened 2023-10-03
Alter MSSQL TPCH Colstore Index Steps

JoshInnis:
Alter the Steps involved in creating colstore indices. Testing the current process in loading data with Colstore indices created before the data is loaded vs after the data is loaded shows a 10-20x improvement when the index is created after loading the data, when BCP is not used. When BCP is used it is more efficient to create the indices before the data is loaded.


sm-shaw (changes requested):
From an overview this looks OK, although I will still plan to test it to be absolutely sure.
One thing that stands out from the overview is that the line 
`set startTime [clock seconds]`
occurs multiple times, however there is no endTime or time report so these lines can be removed.

sm-shaw (approved):

sm-shaw:
I added needed changes with https://github.com/TPC-Council/HammerDB/pull/613/commits/be2c9d3e51040ed2a850e7fd980a7d35e172d272 and have successfully retested

jamiere-msft (approved):

abondvt89 (approved):

PR #614, opened 2023-10-05
TPROCC MSSQLS BCP Load by Warehouse

JoshInnis:
Alter the TPROCC MSSQLS BCP to load the orders and customers per warehouse instead of per district. This results in a faster load time overall.


sm-shaw (approved):
I have tested and verified with consistency checks and all works successfully.

jamiere-msft (approved):

PR #616, opened 2023-10-06
Remove Unneeded Indices From TPROC-H MSSQL 

JoshInnis:
When Column store indices are enabled, the non unique indices on the columns are not required. Removing them will improve the load time of the TPCH load.


sm-shaw (approved):
Have tested and build works correctly

abondvt89 (approved):

jamiere-msft (approved):
I spoke with Josh earlier on this. This matches our TPC-H implementation that we give to our partners for publishing results.

PR #642, opened 2023-11-24
Add optional purge and write back for MariaDB TPROC-C

sm-shaw:
Only compatible with MariaDB 10.7.1 and above where purge_threads are dynamic. 
Functionality is not possible to add to MySQL at the present time as innodb_purge_threads is a read only variable and functionality is dependent on setting purge_threads temporarily to a high value to accelerate the purge process. 

In the CLI the time for purge and write back with require a higher setting of keepalive_time. Plan to add a gdset command to set generic dict settings dynamically rather than only through XML to achieve this. 

Tested for both sync and async options

purge
purge2

### DIFF
--- a/config/mariadb.xml
+++ b/config/mariadb.xml
@@ -40,6 +40,7 @@
             <maria_async_verbose>false</maria_async_verbose>
             <maria_async_delay>1000</maria_async_delay>
             <maria_connect_pool>false</maria_connect_pool>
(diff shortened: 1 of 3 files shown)


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #647, opened 2023-12-05
MSSQLS TPCH Add Flag to Partition Orders and Lineitems

JoshInnis:
Added a flag that allows the Order and Lineitem tables to be partitioned. Each partition spans 1 week. The parition flag is set to false by default and must be enabled.

Column Store and Row store are allowed to be partitioned.

When paritioning is enabled. Certain primary and foreign key definitions defined in HammerDB are no longer valid. HammerDB will still create those keys/constraints when the Orders and Lineitem  tables are not partitioned.`

For columnstore indices, the is an initial degredation of performance, however if the user calls Alter Index ... Reoganize or leaves the server running after the initial load for approximately 48 hours, the performance of power runs improves by 2x when compared to column store indices with no partitioning.

### DIFF
--- a/config/mssqlserver.xml
+++ b/config/mssqlserver.xml
@@ -50,6 +50,7 @@
 	    <mssqls_num_tpch_threads>1</mssqls_num_tpch_threads> 
 	    <mssqls_colstore>false</mssqls_colstore>
 	    <mssqls_tpch_use_bcp>true</mssqls_tpch_use_bcp>
+	    <mssqls_tpch_partition_orders_and_lineitems>false</mssqls_tpch_partition_orders_and_lineitems>
         </schema>
         <driver>
 	    <mssqls_total_querysets>1</mssqls_total_querysets>
--- a/src/mssqlserver/mssqlsolap.tcl
+++ b/src/mssqlserver/mssqlsolap.tcl
@@ -81,148 +81,313 @@ proc CreateDatabase { odbc db azure } {
     }
 }
 
-proc CreateTables { odbc colstore bcp } {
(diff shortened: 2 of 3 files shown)


sm-shaw (changes requested):
Have tested multiple combinations of build and test and all work apart from the following scenario where the refresh function is enabled in the driver script when the partitioning option has been selected.  In this scenario both types of build error with a constraint doesn't exist error. This is when in the refresh function the following is called to allow the refresh function to insert into lineitem first before orders. 

`    $odbc evaldirect "ALTER TABLE LINEITEM NOCHECK CONSTRAINT LINEITEM_ORDER_FK"`
 
The error is the same as follows:
partition orders and lineitem with no clustered columnstore
```

Error in Virtual User 1: Error: [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Constraint 'LINEITEM_ORDER_FK' does not exist.
[Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Could not enable or disable the constraint. See previous errors.
(executing the statement)
```
partition orders and lineitem with clustered columnstore
```
Error in Virtual User 1: Error: [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Constraint 'LINEITEM_ORDER_FK' does not exist.
[Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Could not enable or disable the constraint. See previous errors.
(executing the statement)
```
The fix could be to check if the constraint exists before disabling or re-enabling - or to add a catch statement and ignore failure if the error message from the catch says that the constraint doesn't exist.

JoshInnis:
Updated PR to reflect the issue found.

jamiere-msft (approved):

PR #652, opened 2023-12-29
MSSQL TPCH Create Advanced Statistics

JoshInnis:
Create an option for users loading TPCH for MSSQL to perform a more advanced statistical analysis. Option is set to false by default because of the large increase to load time.


sm-shaw (approved):
Have tested and all works as expected.

Issue #594, opened 2023-09-11
Extend tdbc::odbc interface to add BCP commands for SQL Server schema builds


sm-shaw:
Pull Requests #587 #592 have added bcp based loads for the SQL Server TPROC-C schema, significantly improving build times. Issue #593 proposes the same update for the TPROC-H schema. 

Currently, this feature uses the external bcp command called via exec. It should be possible to extend the tdbc::odbc interface to add the bcp_init, bind and sendrow commands and call these directly from ODBC rather than using the external bcp command

https://learn.microsoft.com/en-us/sql/relational-databases/native-client-odbc-extensions-bulk-copy-functions/bcp-init?view=sql-server-ver16
https://learn.microsoft.com/en-us/sql/relational-databases/native-client-odbc-extensions-bulk-copy-functions/bcp-bind?view=sql-server-ver16
https://learn.microsoft.com/en-us/sql/relational-databases/native-client-odbc-extensions-bulk-copy-functions/bcp-sendrow?view=sql-server-ver16

sm-shaw:
Pull request #653 added. As noted in the PR the long-term solution is to extend the ODBC interface that HammerDB uses to include bcp based commands https://github.com/TPC-Council/HammerDB/issues/594 as using the executable is the cause of issues such as this where the login connection and the bcp connection are out of synch.

JoshInnis:
I have been revisiting this issue, and I have some questions about the details how to approach this.

For extending tdbc::odbc, you mean, pull down the source code and add the bcp options directly into the library? or is it more about using the odbc library that Microsoft developed with this functions in them? Or is there perhaps another route to do this that I have missed?

sm-shaw:
I have had a further look into this and made some additional changes to the tdbc::odbc package to enable it for bcp. Firstly we have to enable for bcp before we connect, and also we need to load an additional library such as msodbcsql17 or msodbcsql18 for the additional functionality (bcp_initA, bcp_exec etc), however I have some concerns about over complexity with having to load the additional libraries. 

So, when you said, _Or is there perhaps another route to do this that I have missed?_ I now think there could be! as, in researching implementing bcp in the library it referenced using bulk insert as a SQL statement instead. This inherits the connection, so we don't need to reconnect to do this. Also, the current bcp implementation doesn't work for entra authentication, however if we inherit the existing connection then I don't see why using bulk insert wouldn't be able to use entra?. 

[...]

PR #653, opened 2024-01-05
Add bcp -u option for SQL Server to trust the server certificate

sm-shaw:
Adds -u option to bcp based builds when running on Linux with ODBC driver 18 and trust server certificate is checked to avoid `certificate verify failed:self signed certificate` error.
Tests on Windows showed bcp used ODBC 17 so change has not been added to Windows but may need to be in future. 

Noting that the long-term solution is to extend the ODBC interface that HammerDB uses to include bcp based commands https://github.com/TPC-Council/HammerDB/issues/594 as issues such as this are a direct result of calling the external tools i.e. we have already logged in and trusted the server certificate, however using the bcp executable means we need to do this again.

### DIFF
--- a/src/mssqlserver/mssqlsolap.tcl
+++ b/src/mssqlserver/mssqlsolap.tcl
@@ -389,7 +389,6 @@ proc CreateDateScheme { odbc } {
         $odbc evaldirect $sql($i)
     }
 }
-
 # bcp command to copy from file to specified tables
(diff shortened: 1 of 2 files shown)


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #658, opened 2024-01-19
Add quotemeta function to escape special characters in passwords

sm-shaw:
Add function based on perl quotemeta https://perldoc.perl.org/functions/quotemeta to escape special characters such as semicolon or brackets in passwords. 
Behaviour is optional with new generic setting of quote_passwords with default of true but if set to false then behaviour will be as previous. 

Depending on the interface some additional manual quoting will be needed as follows:
1. tcl interactive &GUI - no escaping needed
2. Tcl script example: diset connection mssqls_pass =U\#\}r>Y^O~&RcE$:x5:DO[*y5<(a&AH9r2{
3. Python interactive and scripts need to escape square bracket diset('connection','mssqls_pass','=U#}r>Y^O~&RcE$:x5:DO\[*y5<(a&AH9r2{')

At this PR we do not plan to support the change for XML connect pool as this requires further work to escape the characters for the XML parser. 

At the present time the change is designed to be as the least intrusive as possible, especially where special characters are not used in passwords and allow for users who to prefer wrapping passwords to still do so, whereas it will pick up obvious uses and prevent the scripts erroring due to special characters.


abondvt89 (approved):

abondvt89:

Issue #563, opened 2023-06-19
Hope to support oceanbase databases


deskau:
The oceanbase database has two modes: MySQL and Oracle, which are mostly syntax compatible with both MySQL and Oracle. Currently, when testing the oceanbase MySQL mode, there are some issues that we hope to support.
1. For example, in src/mysql/mysqloltp.tcl, the name is timestamp. In the stored procedure required to create a test, there is an input parameter named timestamp. However, in oceanbase, timestamp is a reserved keyword, so creating this stored procedure will result in an error. The current solution is to manually add 'timestamp' for each translation。
2. Because oceanbase does not support the MySQL statistical information collection “analyze table” command, the syntax for reporting errors in buildschema is not supported: it is also necessary to annotate the three lines 488，489，490 in the src/mysql/mysqloltp.tcl file to avoid errors。

sm-shaw:
Hi @deskau - yes this should definitely be possible and something done for other databases by adding a compatible parameter. 
So for example in PostgreSQL we have an Oracle compatible (For EnterpriseDB Advanced Server) and Citus https://github.com/citusdata/citus compatible options that were submitted.
In the dict you see the following: 
```
hammerdb>print dict
Dictionary Settings for PostgreSQL
..
tpcc       {
...
 pg_oracompat        = false
 pg_cituscompat      = false
...
}
```
and this is also shown in the GUI which also sets the correct options such as the use of stored procedures
![image](https://github.com/TPC-Council/HammerDB/assets/38044085/6f8bb7d1-db3d-4819-b764-8546e876b70e)

Then  where needed, so in the build, driver (when modifying the driver changes need to go in multiple sections for the different options such as asynchronous) or both you pass the compatible variable and use this in the script to change the behaviour as required, for example:
```
   if { $citus_compatible eq "true" } {
            set sql(7) { SELECT create_distributed_function('dbms_random(int,int)') }
            set sql(8) { SELECT create_distributed_function(oid, '$1', colocate_with:='warehouse') FROM pg_catalog.pg_proc WHERE proname IN ('neword', 'delivery', 'payment', 'ostat', 'slev') }
        }
```
so if this is of interest then looking at the examples above should be able to help you find out what to do. 

Then you submit a pull request with your changes, for example this was submitted for the Citus Changes https://github.com/TPC-Council/HammerDB/pull/302 (you wouldn't need the additional formatting that was done here) that works for you. 

We can test to make sure everything is still OK with the default MySQL build and test.

[...]

PR #670, opened 2024-02-19
Add Support for oceanbase tpch

yaqi-zhao:
As described in https://github.com/TPC-Council/HammerDB/issues/563. This PR support Oceanbase TPROC-H  by adding a compatible parameter. 

image

Co-authored-by: XiTang <xi.tang@intel.com>

TPROC-H
#Check 1 Database Exist
#Check 2 Tables Exist
#Check 3 supplier count in schema is the same as dict setting
#Check 4 Tables are indexed
#Check 5 Tables are populated
#Consistency check

### DIFF
--- a/config/mysql.xml
+++ b/config/mysql.xml
@@ -12,6 +12,7 @@
         <mysql_ssl_cert>client-cert.pem</mysql_ssl_cert>
         <mysql_ssl_key>client-key.pem</mysql_ssl_key>
         <mysql_ssl_cipher>server</mysql_ssl_cipher>
+        <oceanbase_port>2881</oceanbase_port>        
     </connection>
     <tpcc>
         <schema>
@@ -49,7 +50,10 @@
             <mysql_tpch_pass>mysql</mysql_tpch_pass>
             <mysql_tpch_dbase>tpch</mysql_tpch_dbase>
             <mysql_num_tpch_threads>1</mysql_num_tpch_threads>
-            <mysql_tpch_storage_engine>innodb</mysql_tpch_storage_engine>
+            <mysql_tpch_obcompat>false</mysql_tpch_obcompat>
+            <ob_partition_num>1</ob_partition_num>
+            <ob_tenant_name>tpch</ob_tenant_name>
+            <mysql_tpch_storage_engine>innodb</mysql_tpch_storage_engine>                        
         </schema>
         <driver>
             <mysql_total_querysets>1</mysql_total_querysets>
--- a/config/mysql.xml
+++ b/config/mysql.xml
@@ -12,7 +12,7 @@
         <mysql_ssl_cert>client-cert.pem</mysql_ssl_cert>
         <mysql_ssl_key>client-key.pem</mysql_ssl_key>
         <mysql_ssl_cipher>server</mysql_ssl_cipher>
-        <oceanbase_port>2881</oceanbase_port>        
+        <mysql_oceanbase_port>2881</mysql_oceanbase_port>
     </connection>
     <tpcc>
         <schema>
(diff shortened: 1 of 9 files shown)


sm-shaw:
Many thanks for the pull request. 

I will test to make sure everything is OK with the existing MySQL before approving.

From a quick overview there is a few minor changes to make and I will look to add those to the PR, such as keeping the scale factor and virtual users to build schema at the bottom of the options and also keeping all of the mysql dict options starting with mysql_.

We are currently just about to release v4.10 which has already been tagged and drafted and therefore once merged this is likely to be included with v4.11.

sm-shaw (approved):
I've tested and added an additional commit to make some changes, what I have changed is:

1. Updated some of the variable names for consistency with mysql prefix. 
2. It looks the partitions option is only needed in the build script and not the driver i.e. the variable wasn't used in the driver.  So I've removed the partitions option from the driver and left it in the build.
3. I've moved the fields around for consistency across HammerDB.
4.  The changes to ConnectToMySQL always set the user to oceanbase if the local socket was not used when logging in. I've fixed this so if not using oceanbase and are using a port to connect it still works. 
5.  The transaction counter was changed to work against oceanbase, however the GUI dialog did then not allow an oceanbase option for the transaction counter - I've fixed this by adding options to the transaction counter. 
7. Cloud Analytic queries errored as they didn't include Oceanbase variables in the EDITABLE OPTIONS - this has been fixed. 

I have tested against regular MySQL only to make sure all the TPC-H options work when you don't select the oceanbase options, so am happy to approve on this basis. 

Please check to make sure that everything continues to work on the oceanbase side after the updates.

xtangxtang:
> I've tested and added an additional commit to make some changes, what I have changed is:
> 
> 1. Updated some of the variable names for consistency with mysql prefix.
> 2. It looks the partitions option is only needed in the build script and not the driver i.e. the variable wasn't used in the driver.  So I've removed the partitions option from the driver and left it in the build.
> 3. I've moved the fields around for consistency across HammerDB.
> 4. The changes to ConnectToMySQL always set the user to oceanbase if the local socket was not used when logging in. I've fixed this so if not using oceanbase and are using a port to connect it still works.
> 5. The transaction counter was changed to work against oceanbase, however the GUI dialog did then not allow an oceanbase option for the transaction counter - I've fixed this by adding options to the transaction counter.
> 6. Cloud Analytic queries errored as they didn't include Oceanbase variables in the EDITABLE OPTIONS - this has been fixed.
> 
> I have tested against regular MySQL only to make sure all the TPC-H options work when you don't select the oceanbase options, so am happy to approve on this basis.
> 
> Please check to make sure that everything continues to work on the oceanbase side after the updates.

Thanks Steve, we will check if the oceanbase is corret.  Any process we will update here.

yaqi-zhao:
> Thanks Steve, we will check if the oceanbase is corret

yaqi-zhao:
> I've tested and added an additional commit to make some changes, what I have changed is:
> 
> 1. Updated some of the variable names for consistency with mysql prefix.
> 2. It looks the partitions option is only needed in the build script and not the driver i.e. the variable wasn't used in the driver.  So I've removed the partitions option from the driver and left it in the build.
> 3. I've moved the fields around for consistency across HammerDB.
> 4. The changes to ConnectToMySQL always set the user to oceanbase if the local socket was not used when logging in. I've fixed this so if not using oceanbase and are using a port to connect it still works.
> 5. The transaction counter was changed to work against oceanbase, however the GUI dialog did then not allow an oceanbase option for the transaction counter - I've fixed this by adding options to the transaction counter.
> 6. Cloud Analytic queries errored as they didn't include Oceanbase variables in the EDITABLE OPTIONS - this has been fixed.
> 
> I have tested against regular MySQL only to make sure all the TPC-H options work when you don't select the oceanbase options, so am happy to approve on this basis.
> 
> Please check to make sure that everything continues to work on the oceanbase side after the updates.

Thanks Steve, I have tested the oceanbase with the newest updates, everything continues to work fine.

sm-shaw:
Excellent, once fully approved by all reviewers we will merge the pull request

jamiere-msft (approved):

abondvt89 (approved):

PR #678, opened 2024-03-08
Change GUI properties dailogs to tabbed

sm-shaw:
As we've added options over time, some of the GUI options dialog boxes have become too large. 
This PR takes the typical tabbed properties dialog approach to break the dialog into 2 tabs of connections and settings to reduce the size. 

Example from Linux with MySQL

New Connection
connection
New Settings
settings
Current Dialog
previous


jamiere-msft (approved):

abondvt89:

Issue #679, opened 2024-03-15
Update and Unify SVG UI Themes


sm-shaw:
As discussed in https://github.com/TPC-Council/HammerDB/issues/676 we have the option to update the awthemes https://wiki.tcl-lang.org/page/awthemes and tksvg https://github.com/oehhar/tksvg packages and unify the themes used for both Windows and Linux.

The default will be the awbreeze theme on both platforms changing Linux from awlight with windows keeping the same theme with some updates. awlight will be available as an alternative as well as awarc.  Although dark themes are available at present, there isn't the plan to have a dark theme option. 

awbreeze theme on Linux
![breeze](https://github.com/TPC-Council/HammerDB/assets/38044085/f7c63822-c09c-40c3-9958-76bd8bf48448)

Existing awlight theme on Linux which will remain an option
![awlight](https://github.com/TPC-Council/HammerDB/assets/38044085/aaba5b7f-4e5c-49cc-bb69-a6aa377c131c)

Alternative awarc theme on Linux
![awarc](https://github.com/TPC-Council/HammerDB/assets/38044085/972202a5-1782-432e-b4cd-f924a43b035c)

marjus45:
Hi @sm-shaw , I noticed that these libraries are missing:

* http://www.hammerdb.com/build/InputLibs/awthemes-9.3.1.7z
* http://www.hammerdb.com/build/InputLibs/tksvg-0.5.7z

It seems that it is related with the work you are doing for this issue. Removing these packages from the server breaks building of this package. Is it possible to restate these libraries in the server?

sm-shaw:
Hi, Yes I have been updating the build - progress can be tracked here https://github.com/TPC-Council/HammerDB/compare/master...sm-shaw:HammerDB:679 and this will show how to use the new packages. 

If you want the older ones then let me know and I can restore them - it is great to see people using the build from source feature. 

I was aware that this would break the older builds, however not entirely sure how many people were building from source as building from the older setup files/inputlibs gives you what is already available for pre-compiled download anyway.  So in the past we have just updated the build files to be compatible with the version that is the currently being developed.

marjus45:
@sm-shaw Yes please restore these images as it currently breaks our build.

When the new changes are merged I would be happy to also upgrade the version and try them out.

sm-shaw:
The old files have been restored, and the build tested, so it should all work again as expected.

PR #680, opened 2024-03-21
Update SVG and themes packages, Unify themes on Windows and Linux

sm-shaw:
As discussed in #679 PR updates the svg and themes packages and sets the default theme to be the scalable awbreeze on both Linux and Windows. 
PR also fixes GUI issue where the transaction counter graph does not adjust if the window is resized before it is started.


abondvt89 (approved):

jamiere-msft (approved):

abondvt89:

Issue #681, opened 2024-03-22
Error missing close-brace when selected PostgreSQL XML Connect Pool


sm-shaw:
**Guidance**
Bug reports are for when HammerDB is not behaving as expected.
Bug reports should not be submitted for help in understanding database performance related questions.
General questions on database performance or HammerDB usability should be submitted under Discussions.

**Describe the bug**
When the connection parameters have been configured in the XML file in the connect pool directory and the XML Connect Pool option is selected the workload error

**To Reproduce**
Run the HammerDB PostgreSQL TPROC-C workload and select XML connect pool. 

**Expected behavior**
The workload should run using the connections and policy defined in the XML file. 

**Screenshots**

![missingcb](https://github.com/TPC-Council/HammerDB/assets/38044085/0dfd4ddd-f569-4cbd-8e9a-d715ce67b903)

**HammerDB Version (please complete the following information):**
 - Version: 4.10
 - Build: All

**HammerDB Interface (please complete the following information):**
 - UI: [e.g. GUI or CLI]

**Operating System (please complete the following information):**
ALL

**Database Server (please complete the following information):**
 - Database name: PostgreSQL

**Database Client (please complete the following information):**
 - Database client name: PostgreSQL

**Additional context**
Workaround is to manually add close-brace at end of driver script. 

Note that the XML connect pool option only supports functions and does not support PostgreSQL stored procedures because PostgreSQL stored procedures are not support with prepared statements that is needed to create multiple cursors to use with the different connections.

PR #682, opened 2024-03-22
Fix missing close-brace in PostgreSQL XML Connect Pool option

sm-shaw:
Resolves Issue #681 by modifying lines that are deleted so that final close-brace remains when XML connect pool option is selected. 
Note that as mentioned in the issue PostgreSQL stored procedures are not supported with prepared statements and therefore this option will always default to functions and will error if trying to use stored procedures.


abondvt89 (approved):

jamiere-msft (approved):

abondvt89:

PR #690, opened 2024-04-05
Add Database version to TPROC-C timed test

sm-shaw:
Pull request adds a database version number query to the TPROC-C timed test and prints the string such as `DBVersion:16.0.1000.6` at the start of the test. If the version cannot be queried or the version is not an expected version of digits and dots then the test will not error but instead record the version as `DBVersion:NULL`

This can then be queried at a later point with the jobs db command. 

```
hammerdb>job 6601BCF63E8403E263532363 db
[
 "MSSQLServer",
 "16.0.1000.6"
]
```

The enhancement means we will now have a record of the database version stored along with the test data.

### DIFF
--- a/modules/jobs-1.0.tm
+++ b/modules/jobs-1.0.tm
@@ -1240,7 +1240,18 @@ namespace eval jobs {
                             return
                           } else {
                             if { [ dict keys $paramdict ] eq "jobid db" } {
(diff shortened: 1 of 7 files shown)


abondvt89 (approved):

jamiere-msft (approved):

abondvt89:

PR #701, opened 2024-05-02
Fix incorrect value for SQLite database

sm-shaw:
Change value for SQLite jobs database size from KB to bytes


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #725, opened 2024-07-12
Fixes GUI inline edit errors for MySQL and CLI for MySQL, MariaDB, Oracle, Db2

sm-shaw:
Issue only occurs in all cases when selecting both connect pool and asynch at the same time. 
Verified that SQL Server implementation was not affected by the issue in CLI. 
Issue is CLI index is character based but GUI is line.character therefore -1 moves back 1 line in GUI but only back 1 character in CLI. 
We can identify the index by string is entier and correct the replacement in the GUI by not moving forward a line. 
```
             set syncdrvi(3b) [ expr $syncdrvi(3b) - 1 ]
            #Delete run loop
	      if { [ string is entier $syncdrvi(3b) ] } {
            #CLI
            .ed_mainFrame.mainwin.textFrame.left.text fastdelete $syncdrvi(3a) $syncdrvi(3b)
                } else {
            #GUI
            .ed_mainFrame.mainwin.textFrame.left.text fastdelete $syncdrvi(3a) $syncdrvi(3b)+1l
                }
```


sm-shaw:
Correction to the above in that we can identify the index by string is entier and correct the replacement in the CLI (not GUI) by not moving forward a line.
In the CLI we simulate the text editor that is already built into the GUI with string commands.

jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #730, opened 2024-08-14
Bug Fix, wrong timeout value assignment in xtprof module.

Sheharyar570:
Fixed typo in variable name `xtto` causing incorrect timeout assignment in xtprof module.

### DIFF
--- a/modules/xtprof-1.0.tm
+++ b/modules/xtprof-1.0.tm
@@ -269,7 +269,7 @@ set xtjob_storage 0
 if { [ string is entier $xtgather_timeout ] } { 
 set xtto [expr {$xtgather_timeout * 60}]
 } else {
-set xttto 600
+set xtto 600
 #Set xtgather_timout in case decided to print this value in future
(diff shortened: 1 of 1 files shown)


sm-shaw (approved):
Yes, this is fine.  Note that xtto is usually set from xt_gather_timeout in the configuration that you can see from `print generic`
```
timeprofile          {
 profiler           = xtprof
 xt_unique_log_name = 0
 xt_gather_timeout  = 60
 xt_job_storage     = 1
}
```
So if this value is a valid number xtto will be set to this value * 60 which is the number of minutes to wait for. The clause here is only to catch cases where xt_gather_timeout has been incorrectly set and and in this case sets it to 10 minutes. But of course in this case it would break and therefore setting this fallback value as per this pull request will fix this scenario when xt_gather_timeout is not set to a valid number.

jamiere-msft (approved):

abondvt89 (approved):

PR #748, opened 2024-09-10
Replace bcp executable with bulk insert for SQL Server builds

sm-shaw:
As discussed in https://github.com/TPC-Council/HammerDB/issues/594 using the bcp executable for bulk loads provides excellent performance, however we need to provide login details for bcp that results in extending the function for different connection options. for example on Linux only ODBC driver 18 can trust the server certificate with -u option. 

It would be better if we could use bcp under the existing connection. https://github.com/TPC-Council/HammerDB/issues/594 suggests extending the tdbc::odbc interface to include the bcp_init, bcp_exec commands etc, however bulk insert https://learn.microsoft.com/en-us/sql/t-sql/statements/bulk-insert-transact-sql?view=sql-server-ver16 looks like it already allows us to do this. 

This PR replaces calling the bcp executable with a bulk insert statement instead. Tests show this change further improves load performance.

### DIFF
--- a/src/mssqlserver/mssqlsolap.tcl
+++ b/src/mssqlserver/mssqlsolap.tcl
@@ -125,6 +125,14 @@ exec sp_updatestats
     return
 }
 
+proc location { odbc } {
+set location "remote"
+if ![catch {set rows [ $odbc allrows "select case when cast(serverproperty('machinename') as varchar(100)) = cast(host_name() as varchar(100)) then 'local' else 'remote' end"  ]}] {
+set location [ lindex {*}$rows 1 ]
+	}
+return $location
+}
+
 proc CreateDatabase { odbc db azure } {
     set table_count 0
     puts "CHECKING IF DATABASE $db EXISTS"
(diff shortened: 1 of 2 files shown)


sm-shaw:
Open question for this PR review is whether it will be an issue that bulk insert needs the files on the same host as SQL Server is running as tests have been run locally. 

If it is options are to detect if the file is local or remote and see if adding a servername with UNC is a possible solution? or simply catch the bulk insert and try the original bcp.exe as a backup command if the file is not found - which I was thinking of initially in any case (The same question also applies if we added the bcp commands to the tdbc::odbc interface on where the file is read from after connection). 

From documentation below: 


**data_file must specify a valid path from the server on which SQL Server is running.** If data_file is a remote file, specify the Universal Naming Convention (UNC) name. A UNC name has the form \\SystemName\ShareName\Path\FileName. For example:

Copy
BULK INSERT Sales.Orders
FROM '\\SystemX\DiskZ\Sales\data\orders.dat';
Beginning with SQL Server 2017 (14.x), the data_file can be in Azure Blob Storage. In that case, you need to specify data_source_name option. For an example, see [Import data from a file in Azure Blob Storage](https://learn.microsoft.com/en-us/sql/t-sql/statements/bulk-insert-transact-sql?view=sql-server-ver16&redirectedfrom=MSDN#f-import-data-from-a-file-in-azure-blob-storage).

Azure SQL Database only supports reading from Azure Blob Storage.

jamiere-msft (approved):

sm-shaw:
Also added a -G option if entra is specified for authentication (we don't have the windows username and password so will not be able to add that as a bcp option. 
And in checking all options, bulk insert fails with SQL authentication (username and password) as it does not have permission to read the files so added check so that the location must be local and not SQL authentication otherwise will fall back to bcp.

abondvt89 (approved):

PR #764, opened 2024-11-04
ALTER DATABASE only if necessary

emelsimsek:
CreateUserDatabase returns error in case the user does not have the ALTER DATABASE permissions. It apears that ALTER DATABASE command is not even necessary if the user is already the database owner. So it should be possible to skip this command.


sm-shaw (approved):
Have tested (with PostgreSQL 16.3) and all works fine. Also successfully tested with HammerDB built with Tcl/Tk 9.

jamiere-msft (approved):

abondvt89 (approved):

PR #766, opened 2024-11-05
Free thread shared variables after xtprof run

sm-shaw:
PR unsets dicts and thread shared variables used to gather timing data for the xtprof time profiler after each run and the timing data has been reported. There is also a check to see if the thread shared variables exist at the start of a run which remains. 
Virtual user independent profiler array is already unset after use.


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

Issue #782, opened 2025-04-24
The steprun feature not working as expected.


greenhal:
**Guidance**
Bug reports are for when HammerDB is not behaving as expected.
Bug reports should not be submitted for help in understanding database performance related questions.
General questions on database performance or HammerDB usability should be submitted under Discussions.

**Describe the bug**
When using the steprun function to create a pyramid shape run, e.g. scale up users then scale down users in a stepped manor.     The replicas ignore the duration setting and run until the primary finishes.

**To Reproduce**
Create a steps.xml file where each step has a start_after + duration time of less than the total duration.

`<steps>
        <replica1>
                <start_after_prev>5</start_after_prev>
                <duration>1</duration>
                <virtual_users>5</virtual_users>
        </replica1>
        <replica2>
                <start_after_prev>8</start_after_prev>
                <duration>1</duration>
                <virtual_users>5</virtual_users>
        </replica2>
</steps>`

Then execute the steprun:

`dbset db pg
dbset bm TPC-C
diset tpcc pg_count_ware 100
diset tpcc pg_rampup 0
diset tpcc pg_duration 15
diset tpcc pg_allwarehouse true
diset tpcc pg_raiseerror true
diset tpcc pg_total_iterations 2147483648
diset tpcc pg_partition true
diset tpcc pg_timeprofile false
diset tpcc pg_driver timed
set timeprofile_option false
diset connection pg_host x.x.x.x
diset tpcc pg_superuser xxxxx
diset tpcc pg_superuserpass xxxxxx
diset tpcc pg_storedprocs true
loadscript
vuset vu 6 
steprun
`

The replicas will start the specified number of vu, at the specified delayed time, but they will run until the primary is finished.

**Expected behavior**
I expect the VUs running on the replicas to stop after the duration defined in the steps.xml file.

**Screenshots**
CLI only

**HammerDB Version (please complete the following information):**
 - Version: 4.6
 - Build: build from source

**HammerDB Interface (please complete the following information):**
 - UI: CLI

**Operating System (please complete the following information):**
 - Server OS: Amazon Linux 2
 - Client OS: Amazon Linux 2

**Database Server (please complete the following information):**
 - Database name: PostgreSQL
 - Database Release Version 16.4

**Database Client (please complete the following information):**
 - Database client name: postgres

**Additional context**

sm-shaw:
[...]

PR #787, opened 2025-05-23
Fix steprun replica timing

sm-shaw:
Fix for issue https://github.com/TPC-Council/HammerDB/issues/782 where steprun replicas are not ending when expected. PR changes the opmode from replica to local for steprun workloads so additional timing and completion is done.


jamiere-msft (approved):

abondvt89:

Issue #793, opened 2025-06-16
Copyright notice in output is not UTF-8 and also not valid ASCII


ArneTR:
**Guidance**
Bug reports are for when HammerDB is not behaving as expected. ✅
Bug reports should not be submitted for help in understanding database performance related questions. ✅
General questions on database performance or HammerDB usability should be submitted under Discussions. ✅

**Describe the bug**
In https://github.com/TPC-Council/HammerDB/commit/df8c2d562ad3fac292457e9d621dc473e75ee46d the copyright string was changed to include a copyright symbol.

We use HammerDB and parse the output. The expectation is that this is UTF-8. I believe this is also the intention of HammerDB as UTF-8 is standard nowadays.

The escape sequence `\251` is not valid UTF-8 but is valid only in other encoding (latin-1). It should be `\xc2\xa9`

**To Reproduce**
Parse the output of HammerDB as UTF-8

**Expected behavior**
Output of HammerDB is valid UTF-8 or valid ASCII

**HammerDB Version (please complete the following information):**
 - Version: v5.0
 - Build: Docker Image https://hub.docker.com/layers/tpcorg/hammerdb/latest/images/sha256-41b8f3ebdddc605979f5e5177c0d3c7149987493801e1fc2ee7ef842e6c4466e

**HammerDB Interface (please complete the following information):**
 - UI: CLI

**Operating System (please complete the following information):**
 - Server OS: Ubuntu 22.04
 - Client OS: Ubuntu 22.04

**Database Server (please complete the following information):**
 - Database name: All
 - Database Release Version: Unrelated

**Database Client (please complete the following information):**
 - Database client name: Unrelated

**Additional context**
None

sm-shaw:
Many thanks for raising, the copyright line was getting too long so it was shortened this way. The ASCII character works in the default development environment and compatibility with external tools scripts cannot necessarily be done by default, however it does look like this change so I will test it in both environments to advise on accepting the pull request. 

Thanks again!

PR #794, opened 2025-06-16
Copyright message UTF-8

ArneTR:
Makes the copyright notice valid UTF-8

Closes: https://github.com/TPC-Council/HammerDB/issues/793


ArneTR:
Thanks for the hint.

I was under the impression it would be interpreted as byte sequence.

Moved all the old \251 now to \u00A9. Also in the files you hinted.

jamiere-msft (approved):

Issue #791, opened 2025-06-13
Problem with automated HammerDB test


Jennify16:
Hello, I have installed HammerDB version 5.0 on Windows Server. During the automated HammerDB test, I get the following error message:

Error in Virtual User 1: Failed to load tdbc::odbc - couldn't open "C:/Users/ADMINI~1/AppData/Local/Temp/3/TCL000008
9c/tcl9tdbc1110.dll": permission denied

I'm logged in to the server as administrator and have full access rights to the directory. Could someone help me resolve this issue?

sm-shaw:
Firstly have you tried rebooting the server and does the problem still persist?

With the single file executables it extracts the libraries needed to a temp location rather than using the lib directory that it used before. It looks like this gets written but then cannot be read which on windows could be the writer still has it open. 

If the reboot doesn't work, it could therefore be worth checking whether the disk the TMP/TEMP area is on is slow and try setting an alternative TMP/TEMP environment variable to a different dedicated location before running HammerDB for better performance.

Jennify16:
Hi sm-shaw,
thank you very much. 
I've tried and tested everything you suggested, but the error message still persists. Please see the screenshot below:


The error message only appears for virtual user 1.

sm-shaw:
OK, I have looked into this some more and can reproduce. As a workaround, try increasing the delay and repeat parameters to 1000ms or higher as follows. 

```
hammerdb>vuset delay 1000
hammerdb>vuset repeat 1000
```
or do:

`package require tdbc::odbc` or `librarycheck`

before running the workload at the hammerdb prompt

To recreate I delete the TCL directories in Temp e.g. 

`C:\Users\smsha\AppData\Local\Temp\TCL0000293c`

and set repeat and delay to 0 (in the GUI)

The TCL directory is created at HammerDB startup but does not contain tcl9tdbcodbc1110.dll at this point. 

Running the virtual users in the GUI then causes this error in all of the Virtual Users e.g. 

```
Benchmark Run jobid=6852BF6759E903E293635303
Error in Virtual User 5: Error: Failed to load tdbc::odbc - couldn't open "C:/Users/smsha/AppData/Local/Temp/TCL0000293c/tcl9tdbcodbc1110.dll": permission denied
Error in Virtual User 6: Error: Failed to load tdbc::odbc - couldn't open "C:/Users/smsha/AppData/Local/Temp/TCL0000293c/tcl9tdbcodbc1110.dll": permission denied
[...]

PR #795, opened 2025-06-20
Fix permission denied on loading database libraries on Windows

sm-shaw:
As detailed in #791 a virtual user can report permission denied on Windows when loading a database library. 

This occurs when the virtual user thread attempts to open the library while it is still being created as Windows can only open libraries from a file system. 

Issue does not occur on Linux. 

Fix implements retry logic when loading a database library and if permission denied is received will clear up and retry the load.
 
Testing also showed that on occasions the package require failed but the package commands exist so test also checks if command exists and does the same clean up and retry. 

Tests show that with delay values of 0, libraries now load. 

PR also fixes mismatch of delay values introduced by https://github.com/TPC-Council/HammerDB/commit/5ff2f8aa0c98af733b4281308a4163e6cfee1928

### DIFF
--- a/src/generic/gencli.tcl
+++ b/src/generic/gencli.tcl
@@ -457,35 +457,35 @@ proc vuset { args } {
                 remote_command [ concat vuset vu $val ]
             }
             delay {
-                set conpause $val
-                if { ![string is integer -strict $conpause] } {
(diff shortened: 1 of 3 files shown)


jamiere-msft (approved):

abondvt89:

PR #796, opened 2025-06-20
Add catch to CLI stty command on Linux

sm-shaw:
Adds catch to stty command for terminal reset so error not received if command cannot be found and/or is not needed when running from external script.


abondvt89 (approved):

abondvt89:

PR #797, opened 2025-06-20
Add handling of innodb_snapshot_isolation to MariaDB

sm-shaw:
Testing shows no performance difference with innodb_snapshot_isolation set to 1 or 0 for TPROC-C workload and therefore we can turn it off at the session level to prevent errors being reported when it is set at the global level to 1.


abondvt89 (approved):

abondvt89:

PR #810, opened 2025-07-15
Update Tcl/Tk and TkPath versions

sm-shaw:
Updates HammerDB to Tcl & Tk 9.0.2 which resolves Db2 encoding issues without modifying Tcl/Tk source. 
Also updates tkpath to official repository https://github.com/tcltk-depot/tkpath including HammerDB contributions.
HammerDB version also includes pull request to resolve Canvas text underline issues https://github.com/tcltk-depot/tkpath/pull/9
Note that most of the changes are changes to the back-end packages that are downloaded and built when a build is run and therefore the only update on the HammerDB source side is to change the Tcl version in the Bawt file and the tkpath versions in the Setup file to use the new packages.

### DIFF
--- a/Build/Bawt-2.1.0/Bawt.tcl
+++ b/Build/Bawt-2.1.0/Bawt.tcl
@@ -1344,7 +1344,7 @@ namespace eval BawtBuild {
 
     variable sBuildOpts
     array set sBuildOpts {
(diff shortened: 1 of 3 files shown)


abondvt89 (approved):

abondvt89:

PR #830, opened 2025-10-04
Improve message if metstart fails because sysstat not installed

sm-shaw:
Improves messaging if metstart command fails because agent fails to start when sysstat is not installed.
Add hint to check mpstat command runs. 
Fix not needed on Windows as mpstat included with HammerDB.

### DIFF
--- a/src/generic/genmetrics.tcl
+++ b/src/generic/genmetrics.tcl
@@ -19,7 +19,7 @@ proc ConfigureNetworkDisplay {agentid agenthostname} {
         }
         #Test Agent Port
         namespace import socktest::*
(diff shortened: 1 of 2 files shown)


sm-shaw:
Following message is reported if failure on Linux
```
hammerdb>Connecting to Agent to Display CPU Metrics
Metric receive port open @ 15045 on ubuntu22
Connecting to HammerDB Agent @ localhost:10000
Testing Agent Connectivity...SocketError
Connection failed verify agent hostname and id @ localhost:10000
Check sysstat package is installed and mpstat command is available to agent
Started CPU Metrics for AGENT CONNECTION FAILED:(1 CPUs)
AGENT CONNECTION FAILED
```

abondvt89 (approved):

PR #858, opened 2026-02-17
tidesdb mariadb engine support

guycipher:


guycipher:
Hey!  I'm working on getting TidesDB properly working within MariaDB when running on HammerDB.   This is a rough implementation, feel free to take it apart and ask me questions @sm-shaw if you want.  No expectation to merge, just for analysis.

sm-shaw:
Hi, many thanks for raising this PR and appreciate the work gone into it.  We'd be interested into looking into tidesdb storage engine support, but as an alternative approach is it not possible to set the database configuration values in the database itself via the maria.cnf?
HammerDB already has an option to choose the storage engine and we do in some cases have different options for different engines such as column store. However overall it would appear that for long term maintenance having the engine options in the database configuration file rather than HammerDB would be a better approach. 

```
            <maria_tidesdb_compression>lz4</maria_tidesdb_compression>
            <maria_tidesdb_sync_mode>full</maria_tidesdb_sync_mode>
            <maria_tidesdb_write_buffer_size>134217728</maria_tidesdb_write_buffer_size>
            <maria_tidesdb_bloom_filter>true</maria_tidesdb_bloom_filter>
            <maria_tidesdb_use_btree>false</maria_tidesdb_use_btree>
            <maria_tidesdb_isolation_level>repeatable_read</maria_tidesdb_isolation_level>
            <maria_tidesdb_flush_threads>2</maria_tidesdb_flush_threads>
            <maria_tidesdb_compaction_threads>2</maria_tidesdb_compaction_threads>
            <maria_tidesdb_block_cache_size>268435456</maria_tidesdb_block_cache_size>
```

guycipher:
> Hi, many thanks for raising this PR and appreciate the work gone into it. We'd be interested into looking into tidesdb storage engine support, but as an alternative approach is it not possible to set the database configuration values in the database itself via the maria.cnf? HammerDB already has an option to choose the storage engine and we do in some cases have different options for different engines such as column store. However overall it would appear that for long term maintenance having the engine options in the database configuration file rather than HammerDB would be a better approach.
> 
> ```
>             <maria_tidesdb_compression>lz4</maria_tidesdb_compression>
>             <maria_tidesdb_sync_mode>full</maria_tidesdb_sync_mode>
>             <maria_tidesdb_write_buffer_size>134217728</maria_tidesdb_write_buffer_size>
>             <maria_tidesdb_bloom_filter>true</maria_tidesdb_bloom_filter>
>             <maria_tidesdb_use_btree>false</maria_tidesdb_use_btree>
>             <maria_tidesdb_isolation_level>repeatable_read</maria_tidesdb_isolation_level>
>             <maria_tidesdb_flush_threads>2</maria_tidesdb_flush_threads>
>             <maria_tidesdb_compaction_threads>2</maria_tidesdb_compaction_threads>
>             <maria_tidesdb_block_cache_size>268435456</maria_tidesdb_block_cache_size>
> ```

Hey Steve, no problem at all, very good I'm grateful!

One thing, so in TidesDB you have storage engine level configurations which can be configured globally in the MariaDB server configuration file (maria.cnf), and then you have table options. In TidesDB each table can have its own optional compression algorithm, write buffer size, durability mode, bloom filter settings, and isolation level, these are specified per-table in the CREATE TABLE DDL.

I.e

```sql
CREATE TABLE foo (...) ENGINE = TIDESDB COMPRESSION='LZ4' SYNC_MODE='FULL' WRITE_BUFFER_SIZE=134217728 BLOOM_FILTER=1 ISOLATION_LEVEL='REPEATABLE_READ';
```

These table options cannot be set in maria.cnf since they are part of the CREATE TABLE syntax, so they need to be in HammerDB which generates those statements. This is similar to how ColumnStore is already handled in HammerDB with engine-specific table definitions.

I'm updating the PR to move the global system variables (flush_threads, compaction_threads, block_cache_size) out of HammerDB - those should be configured in maria.cnf as you suggested. Only the per-table CREATE TABLE options remain in HammerDB.

I'm working on some tests locally to assure all is functional.  Thank you Steve.

guycipher:
Hi @sm-shaw, I've updated the code. TidesDB is integrated into HammerDB as an alternative MariaDB storage engine selectable via the maria_storage_engine (TPCC) or maria_tpch_storage_engine (TPCH) configuration parameter in config/mariadb.xml. 

Say when a user selects tidesdb as the engine, the GUI in src/mariadb/mariaopt.tcl reveals a dedicated TidesDB tab exposing all 16 table-level options.

These options are collected into a tidesdb_opts dictionary in the schema build procedures of src/mariadb/mariaoltp.tcl (TPCC) and src/mariadb/mariaolap.tcl (TPCH), which then append them as engine-specific `CREATE TABLE clauses (e.g., COMPRESSION='LZ4' SYNC_MODE='FULL' WRITE_BUFFER_SIZE=134217728 ...)` to every table definition emitted by the CreateTables procedure. 

Engine-wide system variables such as flush threads, compaction threads, and clock block cache size are intentionally omitted from HammerDB and left to configure in maria.cnf as discussed.

I've updated the ChangeLog.  If there are any issues you find or things you'd like different Steve, do let me know.  

Thank you!

guycipher:
I want to add more couple more pieces here,

1. TideSQL(TidesDB MariaDB engine) has a fully automated install script at https://github.com/tidesdb/tidesql, this will install MariaDB with TidesDB included but can be done manually if need be.  The install script can also be passed parameters to customize the automated install.   By default installs latest MariaDB, TidesDB and TideSQL.
2. I've written and continue to write extensive documentation on the design of plugin engine, features and usage on the TidesDB website https://tidesdb.com/reference/tidesql/

Cheers

sm-shaw:
Hi, 
I will need to run tests to make sure it doesn't change existing behaviour. But I see, as you say if options are per table and not global then anything you set per table would need to be in the schema creation, so this is correct. Also if this is the case it would be possible to have different settings depending on what the tables do. e.g. typically WAREHOUSE and DISTRICT benefits from a small block size whereas ORDER_LINE benefits from a larger one. 
Please allow some time as I work through other tasks but I will plan to test this. 
Thanks again!

guycipher:
Hi @sm-shaw no problem at all, take your time.

With that, the warehouse and district differences I mean you can't adjust the low level block sizes, but you can certain thresholds and buffers, compression etc so this can be discussed later, I think default the system will do well for most general workloads based on my own benchmarking.  Also I'm always benchmarking different patterns and workloads at the engine level, right so I'm not too concerned with what you throw at it in regards to types of data, patterns, workloads, etc.  I'm more concerned on how the system will be configured comparing to InnoDB so it's fair.

sm-shaw:
I have tested and the addition is functional, however as the engine still requires the experimental setting I suggest that we moved this pull request to draft and re-evaluate at a point where the engine itself becomes more mature.  This way when it is available for testing it will provide a reliable comparison. 

One suggestion however is that rather than having the configuration settings on the application side as it currently is:

```
Vuser 1:CREATING DATABASE tpcc
Vuser 1:CONFIGURING TIDESDB TABLE OPTIONS
Vuser 1:TidesDB table options: COMPRESSION='LZ4' SYNC_MODE='FULL' WRITE_BUFFER_SIZE=33554432 BLOOM_FILTER=1 USE_BTREE=0 ISOLATION_LEVEL='REPEATABLE_READ' BLOOM_FPR=100 SYNC_INTERVAL_US=128000 LEVEL_SIZE_RATIO=10 MIN_LEVELS=5 SKIP_LIST_MAX_LEVEL=12 SKIP_LIST_PROBABILITY=25 BLOCK_INDEXES=1 KLOG_VALUE_THRESHOLD=4096 MIN_DISK_SPACE=104857600 INDEX_SAMPLE_RATIO=1 BLOCK_INDEX_PREFIX_LEN=16 DIVIDING_LEVEL_OFFSET=2 L1_FILE_COUNT_TRIGGER=4 L0_QUEUE_STALL_THRESHOLD=10
Vuser 1:CREATING TPCC TABLES
```
These could be moved to the maria.cnf as MyRocks does with a rocksdb_default_cf_options e.g. https://smalldatum.blogspot.com/2018/07/default-options-in-myrocks.html. 

Doing this would mean that once the database side is configured to get up and running you would only have to specify the storage engine as TidesDB which is already supported. e.g. RocksDB can be tested in this way.  If you want different settings for different tables RocksDB allows you to specify which column family to use in the comments section. 

https://mariadb.com/docs/server/server-usage/storage-engines/myrocks/myrocks-column-families

If no column family is specified then the default is used. This way there is minimal configuration requirement on the application side and it can all be done on the database server in a similar way to RocksDB.

```
#MyRocks
sql_mode=no_engine_substitution,no_unsigned_subtraction
default-storage-engine=rocksdb
skip-innodb
default-tmp-storage-engine=MyISAM
rocksdb_wal_dir=/opt/rockswal
rocksdb_data_dir=/opt/rocksdata
rocksdb_flush_log_at_trx_commit=0
rocksdb_block_cache_size=64G
rocksdb_max_background_compactions=3
rocksdb_max_background_flushes=1
rocksdb_default_cf_options=level_compaction_dynamic_level_bytes=true;block_based_table_factory={cache_index_and_filter_blocks=1;filter_policy=bloomfilter:10:false;whole_key_filtering=1};optimize_filters_for_hits=true;max_write_buffer_number=4;write_buffer_size=32m;max_bytes_for_level_base=256m;compression_per_level=kNoCompression;bottommost_compression=kLZ4Compression
rocksdb_block_size=8192
rocksdb_cache_index_and_filter_with_high_priority=ON
rocksdb_cache_high_pri_pool_ratio=0.5
```

The additions do look good and work as intended, however doing this on the database side may be more maintainable going forward. 

scr1

sm-shaw (changes requested):
Recommendation to move to draft whilst further development takes place on the engine side. Suggested potential to simply the configuration settings by moving these to server rather than application side in a similar way to RocksDB.

guycipher:
Hi Steve, thank you.  I will review.  RocksDB is way different in how it talks to MySQL and MariaDB.  The plugin engine utilizes TidesDB in which is stable and ties it to MariaDB very tightly.  We continue to extend and stabilize the plugin and are setting up a server just for HammerDB testing of the plugin.   This was done not done fully prior due to unforeseen circumstances thus the integration to HammerDB and configurations aren't set in alignment nor tested to be stable nor profiled.  We can draft the PR no problem whilst everything gets tightened up.  I made this PR to help you get going with TidesDB Steve so you can tell me where I went wrong adding TideSQL to HammerDB. 

Cheers!

sm-shaw:
Alex, thank-you got it, the way that the plugins work don't allow you to do it in the same way as RocksDB. Lets monitor as the engine is developed.

One other thing to note is that by default any error will cause a virtual user to finish, so if you want a Virtual Users to continue you need to set this to false:
`diset tpcc maria_raiseerror false`
Then a Virtual User will ignore the error and continue with the next transaction, it will not retry the same one but it will not abort and reduce the number of overall Virtual Users that you are running.
```
Vuser 43:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 48:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 41:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 46:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 12:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 12:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 12:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 29:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 29:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 17:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 42:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 47:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 12:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 18:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 18:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
Vuser 17:mariaexec/db server: Got error 149 Lock deadlock; Retry transaction during COMMIT
```

guycipher:
Perfect timing, thank you Steve.

guycipher:
Hi @sm-shaw, upon further testing and analysis, I think there is a better approach to benchmarking TidesDB in MariaDB using HammerDB following what you've mentioned.

This is the current flow I've found that works well.

We can simply build MariaDB like normal and config it with TidesDB.  From there you can configure the configuration file sure but you can also use session variables.

Say you want to align InnoDB and TidesDB:
```sql
-- InnoDB disable fsync per commit
SET GLOBAL innodb_flush_log_at_trx_commit = 0;
SET GLOBAL innodb_doublewrite = OFF;

-- TidesDB set defaults for new tables (these apply to all CREATE TABLE ENGINE=TIDESDB)
SET GLOBAL tidesdb_default_compression = 'NONE';
SET GLOBAL tidesdb_default_write_buffer_size = 16777216;  -- 16MB
SET GLOBAL tidesdb_default_bloom_filter = ON;

-- Both with read committed isolation
SET GLOBAL TRANSACTION ISOLATION LEVEL READ COMMITTED;
```

Say for my.cnf
```
plugin_maturity = experimental
plugin_load_add = ha_tidesdb.so
tidesdb_flush_threads = 2
tidesdb_compaction_threads = 2

# is in bytes so if InnoDB buffer/cache is 1gb you'd set below to 1gb in bytes
tidesdb_block_cache_size =  1073741824 
tidesdb_max_open_sstables = 256
tidesdb_log_level = WARN

```



I'm able to build a schema easily like so:
```sql
dbset db maria
dbset bm TPC-C
diset connection maria_host 127.0.0.1
diset connection maria_port 3306
diset connection maria_socket /tmp/mariadb.sock
diset tpcc maria_user hammerdb
diset tpcc maria_pass hammerdb
diset tpcc maria_dbase tpcc_tidesdb
diset tpcc maria_storage_engine tidesdb
diset tpcc maria_count_ware 10
diset tpcc maria_num_vu 4
diset tpcc maria_partition false
buildschema
```

One thing that would be good Steve is a way to pass table options through CLI for workflows, as TidesDB works in this way.

If you needed to run sync off, you have to do this through an alter table like so:
```
ALTER TABLE warehouse   SYNC_MODE='NONE';
ALTER TABLE district    SYNC_MODE='NONE';
ALTER TABLE customer    SYNC_MODE='NONE';
ALTER TABLE history     SYNC_MODE='NONE';
ALTER TABLE orders      SYNC_MODE='NONE';
ALTER TABLE new_order   SYNC_MODE='NONE';
ALTER TABLE order_line  SYNC_MODE='NONE';
ALTER TABLE stock       SYNC_MODE='NONE';
ALTER TABLE item        SYNC_MODE='NONE';
```

There is INTERVAL mode as well which I say to use if InnoDB is full durability, TidesDB syncs every write so to be fair, INTERVAL mode should be used.

With that, I think we can probably close the PR and add general features to HammerDB, like the table option feature which for example sysbench has this capability which is very useful!!

RR does conflict I see in HammerDB but other isolation levels like read committed, snapshot, etc don't seem to.  

Anyway, thank you Steve!

PR #861, opened 2026-04-07
Remove arbitrary upper limit on num_vu for schema build

hqakhtar:
Remove the hardcoded upper limit on the number of virtual users for schema building across all database backends. The warehouse/scale-factor count constraint already provides a natural upper bound, making the arbitrary numeric cap redundant. This will help speedup the data population by utilizing a bigger set of workers.

Changes:
- verify_build_threads: remove maximum parameter, validate positive integer only
- gencli.tcl: remove 1024 cap from TPC-C and TPC-H CLI validation
- All database spinboxes: set upper bound to 100000 (matching max warehouse count)
- Update all verify_build_threads callers to drop the third argument

### DIFF
--- a/src/db2/db2opt.tcl
+++ b/src/db2/db2opt.tcl
@@ -498,7 +498,7 @@ proc configdb2tpcc {option} {
         "default" {
             ttk::button $Name -command {
                 set db2_count_ware [ verify_warehouse $db2_count_ware 100000 ]
-                set db2_num_vu [ verify_build_threads $db2_num_vu $db2_count_ware 1024 ]
+                set db2_num_vu [ verify_build_threads $db2_num_vu $db2_count_ware ]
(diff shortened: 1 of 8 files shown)


sm-shaw:
Hi, many thanks for the pull request. Yes this looks very reasonable. The limit was an artifact from a time that systems/core counts were not as performant as today so as a concept this is fine to remove this.
From an overview it looks fine, I will test before approval.
The next TPC-OSS meeting is 14th April around when all the current pull requests are expected to be merged.

sm-shaw (approved):
Reviewed and approved to remove build thread limit.

jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #863, opened 2026-04-09
Add Active Session History (ASH) metrics viewer for MySQL and MariaDB

### LINKED ISSUE #842: Add Active Session History Metrics viewer for MariaDB & MySQL
**Is your feature request related to a problem? Please describe.**
HammerDB has an active session history metrics viewer for Oracle & PostgreSQL as shown below, requested to add same solution for MariaDB/MySQL to view performance schema metrics during a workload.
Image

**Describe the solution you'd like**
The same solution as Oracle & PostgreSQL for MariaDB/MySQL

**Describe alternatives you've considered**
N/A

**Additional context**
Equivalent to pg_sentinel in PostgreSQL is required for MariaDB/MySQL and example is here:
https://github.com/hatem-mahmoud/Ash_sampler
https://mahmoudhatem.wordpress.com/2019/03/28/mysql-ash_sampler-a-simple-ash-builder/

crprashant:
Summary

Implements #842: Adds Active Session History (ASH) metrics viewer for MySQL and MariaDB, adapting the existing PostgreSQL ASH viewer (pgmet.tcl).

Changes

New Files

 - src/mysql/mysqlmet.tcl — Full MySQL ASH metrics implementation (~2900 lines)
 - src/mariadb/mariamet.tcl — Full MariaDB ASH metrics implementation (~2900 lines)

Modified Files

 - src/generic/genmetrics.tcl — MySQL/MariaDB namespace dispatch and cleanup handlers

Features

 - Real-time Active Session History chart with stacked wait event bars
 - Wait event type mapping from performance_schema paths to short categories (CPU, Lock, Mutex, RWLock, SXLock, Table_IO, File_IO, Network, Cond, Idle)
 - SQL_ID detail panel with digest hash, %Total_DB_Time, SQL_TYPE
 - SQL text, SQL I/O, and SQL stats tabs
 - Session and event breakdown panels
 - SQL Over Time load chart per digest
 - ASH sampler via event scheduler sampling performance_schema.threads, events_waits_current, and events_statements_current
 - Automatic hammerdb_ash schema and active_session_history table creation
 - Auto-enables required Performance Schema consumers (events_waits_current, events_statements_current) and instruments (wait/%, statement/%) on startup — no manual setup required
 - Synthetic digest generation via SHA2(sql_text, 256) for statements with NULL native digest (common inside stored procedures)
 - SQL stats fallback: tries events_statements_summary_by_digest first, falls back to ASH sample count for synthetic digests

Technical Details

 - Uses mysqltcl/mariatcl APIs with hash-ordered column output matching the pg_select array iteration pattern
 - Julian day constant: PostgreSQL 2450000 → MySQL/MariaDB 719528 (= TO_DAYS('1970-01-01'))
 - MySQL digest (VARCHAR hex) handled as string vs PostgreSQL queryid (bigint)
 - Cleanup handlers wrapped in catch blocks for safe database switching

Testing

 - MySQL
  8.0.45 — native install, Ubuntu 24.04 Azure VM
 - MariaDB
  11.8.6 — Docker container, Ubuntu 24.04 Azure VM
 - Windows 11 HammerDB client, Bawt build (mysqltcl
  3.052.1, mariatcl 0.1.1)
 - TPC-C workload, 2–5 virtual users, verified:
  - ASH chart rendering with colored wait event bars
  - SQL_ID panel with multiple digests (native + synthetic)
  - SQL text showing stored procedure internals (CALL NEWORD, UPDATE stock, etc.)
  - SQL stats populated for both native and synthetic digests
  - Events and sessions panels
  - Clean logs with no errors

Screenshots

MySQL ASH Metrics:

MariaDB ASH Metrics:

Closes #842

## Screenshots
_Screenshots of MySQL and MariaDB ASH metrics in action will be added in comments below._
image
image
image
image
image
image

Closes #842


crprashant:
Hi @abondvt89 , @jamiere-msft, @sm-shaw please provide the feedback. Thanks!

sm-shaw:
Thank-you for the pull request. This looks good. Please bear with us as we have a number of additional pull requests to make and merge first but this will make the next release 5.1 so there is no urgency.

sm-shaw:
I tested the ASH implementation on both MariaDB and MySQL using TPROC-C this is nearly there but needs some further updates. The sampling and graphing are working, but the SQL panels are largely empty or incomplete. Stepping through both engines shows that this is a combination of setup requirements and over-reliance on digest-based grouping.

MariaDB

With:

performance_schema=ON
event_scheduler=ON
performance-schema-instrument='wait/%=ON'
performance-schema-instrument='statement/%=ON'

running a TPROC-C workload populates hammerdb_ash.active_session_history, but initially all SQL-related fields are NULL:

| ash_id | ash_time                   | thread_id | processlist_id | user | host      | db   | command | state     | wait_event_type | wait_event | sql_text | digest_text | digest |
|--------|----------------------------|-----------|----------------|------|-----------|------|---------|-----------|-----------------|------------|----------|-------------|--------|
|      1 | 2026-04-12 07:55:09.588966 |        19 |              5 | root | localhost | tpcc | Query   | Update    | CPU             | CPU        | NULL     | NULL        | NULL   |
|      2 | 2026-04-12 07:55:09.588966 |        27 |              7 | root | localhost | tpcc | Query   | Update    | CPU             | CPU        | NULL     | NULL        | NULL   |
|      3 | 2026-04-12 07:55:09.588966 |        30 |              8 | root | localhost | tpcc | Query   | starting  | CPU             | CPU        | NULL     | NULL        | NULL   |

At the same time, the ASH graph renders partially, so sampling, wait classification, and the scheduler/event loop are all working, but the SQL panels are blank.

mariadbfeed

Checking Performance Schema consumers shows:

SELECT NAME, ENABLED
FROM performance_schema.setup_consumers
WHERE NAME IN (
  'global_instrumentation',
  'thread_instrumentation',
  'events_waits_current',
  'events_statements_current',
  'statements_digest'
);
| NAME                      | ENABLED |
|---------------------------|---------|
| events_statements_current | NO      |
| events_waits_current      | NO      |
| global_instrumentation    | YES     |
| thread_instrumentation    | YES     |
| statements_digest         | YES     |

So the implementation is querying tables that are not collecting data. After enabling them:

UPDATE performance_schema.setup_consumers
SET ENABLED='YES'
WHERE NAME IN (
  'events_waits_current',
  'events_statements_current'
);
| events_statements_current | YES |
| events_waits_current      | YES |

SQL text starts to appear in ASH:

| wait_event_type                | sql_text                                                     | digest_text | digest |
|--------------------------------|--------------------------------------------------------------|-------------|--------|
| wait/io/table/sql/handler      | CALL NEWORD(10,20,1,2493,15,@disc,@last,@credit,...)         | NULL        | NULL   |
| wait/io/table/sql/handler      | UPDATE stock SET s_quantity = no_s_quantity WHERE ...        | NULL        | NULL   |

This confirms that the join to events_statements_current is working and SQL text is being captured. However, digest and digest_text remain mostly NULL, especially for TPROC-C stored procedure activity and internal statements executed inside the procedures.

mariadbfeed2


So it looks like on MariaDB:

ASH sampling works
wait/event graph works
SQL text capture works after enabling the missing consumers
digest-based grouping is still not reliable

MySQL

MySQL also needs additional setup same as MariaDB. 

Initial check:

SELECT NAME, ENABLED
FROM performance_schema.setup_consumers
WHERE NAME IN (
  'global_instrumentation',
  'thread_instrumentation',
  'events_waits_current',
  'events_statements_current',
  'statements_digest'
);
| NAME                      | ENABLED |
|---------------------------|---------|
| events_statements_current | YES     |
| events_waits_current      | NO      |
| global_instrumentation    | YES     |
| statements_digest         | YES     |
| thread_instrumentation    | YES     |

So MySQL is not fully ready out of the box either. events_waits_current also needs to be enabled:

UPDATE performance_schema.setup_consumers
SET ENABLED='YES'
WHERE NAME = 'events_waits_current';

Once configured, MySQL clearly captures real workload SQL in active_session_history, including:

CALL NEWORD(...)
CALL PAYMENT(...)
CALL SLEV(...)
SELECT ... INTO ...
UPDATE ...
START TRANSACTION

myslqfeed2

However, digest coverage also looks incomplete with the same empty panels. 

A direct count shows:

SELECT
  CASE WHEN digest IS NULL THEN 'NULL digest' ELSE 'Has digest' END AS digest_status,
  COUNT(*) AS rows_count
FROM hammerdb_ash.active_session_history
GROUP BY digest_status;
| digest_status | rows_count |
|---------------|------------|
| NULL digest   | 5755       |
| Has digest    | 4111       |

So even on MySQL, a large proportion of sampled rows have no digest.

Grouping by digest confirms this:

SELECT digest, digest_text, COUNT(*) AS total
FROM hammerdb_ash.active_session_history
GROUP BY digest, digest_text
ORDER BY total DESC
LIMIT 20;

The top result is:

NULL / NULL = 5895 rows

followed by top-level stored procedure calls such as CALL NEWORD(...), CALL PAYMENT(...), CALL SLEV(...), CALL DELIVERY(...), and CALL OSTAT(...).

So on MySQL:

ASH sampling works
wait/event graph works
SQL text capture works
digest-based grouping is only partial
Overall conclusion

Both MariaDB and MySQL require additional setup before the feature works properly:

MariaDB needs events_waits_current and events_statements_current enabled
MySQL also needs events_waits_current enabled

After that setup, the sampler works on both engines and real workload SQL is captured.

The remaining issue is that the implementation depends too heavily on digest-based grouping. On both MySQL and MariaDB, many statements, especially those executed inside stored procedures, have NULL digest. As a result:

the graph works
Some SQL text is present
the SQL/event panels remain empty
Summary
ASH sampling: working
wait/event graph: working
SQL text capture: working
both engines require additional Performance Schema consumer setup
digest-based grouping: incomplete on both engines
Recommendation
Ensure the required consumers are enabled for both MySQL and MariaDB.
Do not rely solely on digest for grouping.

A more robust approach would be to fall back to sql_text when digest IS NULL, or combine digest grouping with raw SQL text.

sm-shaw (changes requested):
As per comments in PR. Both MariaDB and MySQL need additional setup steps to begin showing data. Once running the chart is populated, however the SQL and event panes all remain blank. Appears to be digest related. Tested on MariaDB 12.3.0 and MySQL 8.4.4 on Linux Ubuntu 24.

crprashant:
Thanks for the feedback, will address this in next iteration

crprashant:
@sm-shaw Thank you for the thorough testing and detailed feedback. I've addressed all the issues in commit c453027.
 
 ### 1. Performance Schema Consumer Auto-Setup
 
 Both MySQL and MariaDB now automatically enable the required consumers on ASH startup:
 
 UPDATE performance_schema.setup_consumers SET ENABLED='YES'
   WHERE NAME IN ('events_waits_current', 'events_statements_current');
 UPDATE performance_schema.setup_instruments SET ENABLED='YES', TIMED='YES'
   WHERE NAME LIKE 'wait/%';
 UPDATE performance_schema.setup_instruments SET ENABLED='YES', TIMED='YES'
   WHERE NAME LIKE 'statement/%';

Wrapped in catch blocks so it fails gracefully if the user lacks UPDATE privileges. No manual setup required.

2. Synthetic Digest for NULL Digest Rows

The ASH sampler now generates a synthetic digest using SHA2(sql_text, 256) when the native digest is NULL. This is common for statements inside stored procedures (CALL NEWORD, UPDATE
stock, SELECT ... FROM item, etc.). The digest_text field also falls back to sql_text via COALESCE.

3. SQL Stats Fallback

The sql stats query now uses UNION ALL — first tries events_statements_summary_by_digest (full stats for native digests), then falls back to ASH sample count for synthetic digests.

Results After Fixes

MySQL 8.0.45:








MariaDB 11.8.6:




All panels now populate correctly on both engines without any manual Performance Schema configuration.

sm-shaw:
As additional feedback, I double checked I'm using the updated version but still have the same issue with the blank panels. My test versions are MariaDB 12.3.1 and MySQL 8.4.4. Also sometimes I still see sql_text, digest_text and digest as NULL. 

One thought - when porting to HammerDB 5.0 I had an issue with the Oracle Active Session History not populating some of the panels with the commit in particular here: 

**Fix issue with Oracle metrics showing a blank display** https://github.com/TPC-Council/HammerDB/commit/d3e8088db9f673bccff460049efa26401afc88c5

If it is working for you and not for me - perhaps ash_fetch is failing in a similar way? It was missing an event group in this case scheduler. So perhaps the updated versions have groups that are not handled. 

maria_ash
mysql_ash

```
5
*************************** 3452. row ***************************
         ash_id: 4721
       ash_time: 2026-04-14 20:00:45.377939
      thread_id: 59
 processlist_id: 12
           user: root
           host: localhost
             db: tpcc
        command: Query
          state: updating
wait_event_type: wait/synch/mutex/sql/THD::LOCK_query_plan
     wait_event: wait/synch/mutex/sql/THD::LOCK_query_plan
       sql_text: NULL
    digest_text: NULL
         digest: NULL


*************************** 828. row ***************************
         ash_id: 1125
       ash_time: 2026-04-14 20:21:05.000952
      thread_id: 1092
 processlist_id: 920
           user: root
           host: localhost
             db: tpcc
        command: Query
          state: starting
wait_event_type: wait/lock/table/sql/handler
     wait_event: wait/lock/table/sql/handler
       sql_text: NULL
    digest_text: NULL
         digest: NULL

```

crprashant:
@sm-shaw Thank you for the detailed testing and feedback on MySQL 8.4.4 and MariaDB 12.3.1.
 
 I've pushed commit cbfdefe addressing the blank panel issues:
 
 Fixes applied (both mysqlmet.tcl and mariamet.tcl):
 
 
  1. "Other" wait group — Added grey-colored "Other" group mirroring the Oracle fix (d3e8088). Unclassified wait events now fall through to "Other" instead of "CPU"
  2. "Other" computed in fetch procs — To avoid Tcl hash-order column issues with mysql_all/maria_all, "Other" is calculated as Total - sum(known groups) in the fetch procs rather than
 as a SQL column
  3. Event panel %Total_Time — Classified raw wait_event_type values to display group names so createevtFrame can look up the correct color and render bars
  4. SQL stats matching — Fixed digest lookup to use dual matching (DIGEST + DIGEST_TEXT) against events_statements_summary_by_digest. Removed SHA2 fallback in ASH collection to store 
only native DIGEST for consistent matching
  5. SQL stats auto-refresh — Stats panel now updates when clicking different SQL_ID rows while the stats tab is active
  6. Empty timeline selection — Detail panels now clear when selecting empty ranges in the ASH chart
  7. sqlovertimeload query — Fixed column order and added CASE classification with "Other" fallback
 
 Tested on: MySQL 8.4.8 (port 3306) and MariaDB 12.3.1 RC (port 3307) with TPC-C workloads. ASH chart, SQL_ID panel, event panel, and sessions panel all display correctly.
 
 Note: SQL stats (rows_examined, exec_time, etc.) have limited data for TPC-C workloads because MariaDB doesn't track individual SQL statements inside stored procedures in 
 events_statements_summary_by_digest — it only tracks the outer CALL statements. This is a MariaDB architectural difference from Oracle/PostgreSQL. The ASH sample count (calls) is
 accurate.
image
image

sm-shaw:
I still have the same problem under testing.  I can see the details in the right hand pane, however the smaller ones sql_tbl, ses_tbl, evt_tbl continue to remain blank in both MySQL and MariaDB testing with MySQL 8.4.4 and MariaDB 12.3.1 on Ubuntu 24. 

Can you please attach your my.cnf/maria.cnf in case there is a configuration setting that is causing the problem with the blank panels as for now I cannot get these panels to populate.

image

crprashant:
Hi @sm-shaw — thanks for retesting and for the detailed feedback. The blank panels turned out to be two bugs in the patch rather than a config issue, so no my.cnf / mariadb.cnf changes
should be required. Both are now fixed in the latest push (9763bd8).

1. Timezone-dependent timestamp math (blank sql_tbl / evt_tbl / ses_tbl)

The drill-down time filter was computing:

 (TO_DAYS(ash_time) - 719528) * 86400 + TIME_TO_SEC(ash_time)

and comparing it against the drag-selected sample_id range. TO_DAYS / TIME_TO_SEC operate on the value as-is (session-local), while sample_id is a UTC epoch — so on any non-UTC server
the derived value is shifted by the session offset and the WHERE range lands outside the sampled window, returning zero rows. Ubuntu 24 defaults to system-local time for the MySQL /
MariaDB session, which is why you saw it and I didn't on my UTC test box.

Rewrote both ashtime_fetch and ash_details to compare directly against sample_id (and widen the lower bound by one second to include the edge sample), which removes the TZ dependency
entirely.

2. SQL-text panel not restoring after clicking an event row

The right-hand output widget is shared between the sql-text view and the event-info view. Clicking an event row in evt_tbl correctly overwrote it with event details, but clicking the 
sql text button afterwards only dispatched an async mon_execute ash_sqltxt — and if the thread_actv lock was contended, or the digest's sql_text was NULL (common for Connect command
rows on MySQL 8.4, and for some sampled Query rows), the fetch callback either didn't fire or cleared the widget without reinserting anything. The user was left staring at stale event
info and assumed the panel was broken.

Fixed by:

 - Clearing the widget synchronously in ash_sqltxt and showing a -- Loading SQL text ... -- placeholder before dispatching the query.
 - Tracking inserted-row count in ash_sqltxt_fetch and displaying -- No SQL text recorded for this digest -- when the result set has no non-empty sql_text, so NULL-sql_text digests 
give unambiguous feedback instead of looking blank.

Applied identically to src/mysql/mysqlmet.tcl and src/mariadb/mariamet.tcl.

Validation

Retested locally on MySQL 8.4 / Ubuntu-equivalent session TZ. After a TPROC-C run:

 - Drag-select on the ASH chart → sql_tbl, evt_tbl, ses_tbl all populate.
 - Click a sql_tbl row → SQL text appears on the right (or the fallback message for NULL digests).
 - Click an event row → event info replaces it.
 - Click the sql text button again → SQL text is restored cleanly.

Could you give the latest commit a spin on your MySQL 8.4.4 / MariaDB 12.3.1 Ubuntu 24 setup with your existing configs? Happy to compare my.cnf / mariadb.cnf if anything still doesn't
populate, but I'm hopeful the TZ fix alone covers your environment.

sm-shaw:
Excellent, this is looking really good, that fixed the issue with the blank panels. 
There is one more minor fix needed which was inherited from the PostgreSQL metrics and I have fixed here https://github.com/TPC-Council/HammerDB/pull/870
If performance schema is not enabled then it does not exit gracefully, applying the same fix as PostgreSQL should fix this and is worth adding. 
```
Database Metrics Error: performance_schema is not enabled on this MariaDB instance
Database Metrics Error: 
Metrics Connected
call mon_execute error:thread "tid0x742cd91fe6c0" does not exist
```
This is the output I now get which is great. 
mariadb_ash
mysql_ash
One comment/observation is that the biggest table_IO event tends to be inserting into hammerdb ash itself. Therefore there may be a way to optimize the SQL to reduce the overhead. 
top_sql
 However this is a nice to have and with with the performance schema not enabled adding a graceful exit it would be good to go and approve ASAP.

sm-shaw:
One quick suggestion to resolve the ash query showing top in the list of events is to exclude any query on `hammerdb_ash.active_session_history` in the ash output.

sm-shaw:
Another suggestion to reduce the impact is that currently we are refreshing the hammerdb metrics every 15 seconds 
` set public(ash,load_bucket_secs) 15`
So if the ash_sampler could be reduced from every second to for example every 5 seconds to reduce the impact.
```
CREATE EVENT IF NOT EXISTS hammerdb_ash.ash_sampler
                                ON SCHEDULE EVERY 1 SECOND
DO    
```      
Doing this and excluding any queries on hammerdb_ash would make the impact of the query unnoticeable to the user.

crprashant:
Thanks for the review — I've pushed `4296c19` which addresses all three points from your latest comments. Identical changes applied to `src/mysql/mysqlmet.tcl` and `src/mariadb/mariamet.tcl`.

**1. Graceful exit when `performance_schema` is disabled**

`connect_to_mysql` / `connect_to_maria` now probe `SHOW VARIABLES LIKE 'performance_schema'` before creating `hammerdb_ash`. When OFF, we emit a clear message, close the handle, and signal the metrics thread to quit — no more "thread does not exist" noise:

```
Database Metrics Error: performance_schema is not enabled on this MySQL/MariaDB instance.
Enable with performance_schema=ON in my.cnf/mariadb.cnf and restart the server.
Metrics Connection Failed: Verify Metrics Options
```

**2. Exclude `hammerdb_ash` activity from ASH output**

The sampler now filters out any active statement touching the ASH schema, so the `INSERT INTO hammerdb_ash.active_session_history …` from the event scheduler *and* the metrics-thread `SELECT … FROM hammerdb_ash…` reads are never recorded in the first place:

```sql
...
LEFT JOIN performance_schema.events_statements_current s
  ON t.THREAD_ID = s.THREAD_ID
WHERE t.TYPE = 'FOREGROUND'
  AND t.PROCESSLIST_COMMAND NOT IN ('Sleep','Daemon')
  AND t.PROCESSLIST_ID IS NOT NULL
  AND (s.SQL_TEXT IS NULL OR s.SQL_TEXT NOT LIKE '%hammerdb_ash%')
```

This removes the self-inflicted `wait/io/table/sql/handler` entries from the top events / SQL_ID panels that were showing ASH as the heaviest Table_IO consumer.

**3. Slower sampler to reduce observer overhead**

Schedule reduced from `EVERY 1 SECOND` to `EVERY 5 SECOND`, and I added `DROP EVENT IF EXISTS hammerdb_ash.ash_sampler` before the `CREATE EVENT IF NOT EXISTS` so that existing `hammerdb_ash` databases pick up the new cadence and filter instead of reusing the original event:

```sql
DROP EVENT IF EXISTS hammerdb_ash.ash_sampler;
CREATE EVENT IF NOT EXISTS hammerdb_ash.ash_sampler
  ON SCHEDULE EVERY 5 SECOND
  DO
  INSERT INTO hammerdb_ash.active_session_history ( ... ) SELECT ...;
```

With `load_bucket_secs=15` on the read side and the sampler at 5 s, the combined footprint on the workload is effectively unnoticeable.

**Validation**

Re-tested end-to-end on MySQL 8.4.4 (port 3306) and MariaDB 12.3.1 (port 3307) on Ubuntu 24, with TPROC-C running:
- `performance_schema=ON` → all four panels (`sql_tbl` / `evt_tbl` / `ses_tbl` / right-hand SQL stats) populate with real TPC-C digests; no `hammerdb_ash` entries appear in the top events.
- `performance_schema=OFF` → clean early exit with the message above; GUI returns control instead of hanging.

*Config files used for the re-test (in case they help):*

MySQL — `/../.cnf`
```ini
[mysqld]
performance_schema = ON
performance-schema-instrument = 'wait/%=ON'
performance-schema-instrument = 'statement/%=ON'
performance-schema-consumer-events-waits-current = ON
performance-schema-consumer-events-statements-current = ON
```

MariaDB —  started with `--performance-schema=ON`; `/../.cnf` defaults plus `performance_schema = ON` under `[mysqld]`.

Please give `4296c19` a spin — happy to tweak further if anything still looks off.

sm-shaw:
Many thanks for the revisions. I have tested and yes the error message is now fixed the same as the PostgreSQL one.
```
Waiting for Connection to MariaDB for Database Metrics...
Metrics Connecting to host:127.0.0.1 port:3306
Database Metrics Error: performance_schema is not enabled on this MariaDB instance. Enable with performance_schema=ON in my.cnf/mariadb.cnf and restart the server.
Metrics Connection Failed: Verify Metrics Options
```
and the display now shows correctly. 
maria_ash2
mysql_ash2
Happy to approve 😄

sm-shaw (approved):
Approve after changes and retest.

jamiere-msft (approved):

abondvt89 (approved):

abondvt89:
@crprashant Thank you for all your efforts and responsiveness to get this PR in merge shape.

PR #865, opened 2026-04-10
Standardise Cluster Connect Pool terminology

### LINKED ISSUE #828: Change XML Connect Pool option to Cluster Connect
**Is your feature request related to a problem? Please describe.**
The XML Connect Pool option could be named better as the pool of connections is nearly always used for cluster testing with multiple endpoints. 

**Describe the solution you'd like**
Change the naming under the GUI and CLI from connect pool to cluster connect. eg mysql_connect_pool dict setting would be mysql_cluster_connect. 

**Describe alternatives you've considered**
An additional option is to also include a user configurable location for the cluster connect file. The default location would be the config/clusterconnect directory, however the GUI & CLI would include an additional option. In the GUI this is grey out until cluster connect is selected. This then makes the option much clearer that the user needs to define multiple end-point connections in a separate file. 

**Additional context**
The additional option would go at the end as shown here. 
Image
Update should be for all databases.

sm-shaw:
This change standardises the user-facing terminology to “Cluster Connect Pool” across the UI and supporting code to highlight that typical usage is for clusters.

### DIFF
--- a/modules/tpcccommon-1.0.tm
+++ b/modules/tpcccommon-1.0.tm
@@ -174,7 +174,7 @@ namespace eval tpcccommon {
       async_time [ expr $as_thkt * 1000 ]
     }
   }
-  #Find a Valid XML Connect Pool Config Directory,don't look in zipped fs as user needs to edit
+  #Find a Valid Cluster Connect Pool Config Directory,don't look in zipped fs as user needs to edit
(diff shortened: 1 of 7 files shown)


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #866, opened 2026-04-11
Fix Windows clearline in default COLOR

### LINKED ISSUE #779: On Windows HammerDB v5.0 CLI will only clear next line in black colour
In earlier versions of HammerDB, the CLI is started with a batch file that includes the command: COLOR 07 to set the terminal to a black background, which is the default. HammerDB v5.0 does not use a batch file as it uses executables. Therefore, if a non-default terminal background (e.g. have started a command window and are calling hammerdb by typing the command rather than double-click) is used with HammerDB v5.0 then when the line is cleared for the prompt, it is always cleared in the default colour ie black: 

Image

The line is cleared to make sure there is no other text on the prompt line. 

The command used is the fill_console command on Windows with the Linux alternative as shown:
There is no option to change the fill color so it remains at the system default which is COLOR 07.
```
  #col always 9 as length of hammerdb> prompt
        twapi::fill_console $handle -numlines 1 -position "9,$row"
    } else {
        print "[ESC]\[2K\r" nowait
```
As a command is typed it will fill it in the correct console color, therefore this does not affect functionality. If the default COLOR 07 background is used, then it will also not be seen. 

As a potential fix, if we comment out the fill_console command then the line is not cleared and the colour not shown, however needs further testing and as the terminal currently works as expected and as the issue is not functional we will wait for further testing to ensure that not filling the next line has no adverse impact. 

The straightforward workaround is to type "COLOR 07" before calling HammerDB to set the console background:

Image

sm-shaw:
Fixes issue with Windows console only clearing the line in black. 
Thanks to Ashok Nadkarni for pointer to Twapi command `set attrs [twapi::get_console_screen_buffer_info $handle -textattr]`
Fixed for both Tcl and Python modes.
image

### DIFF
--- a/modules/tclreadline-1.2.1.tm
+++ b/modules/tclreadline-1.2.1.tm
@@ -113,11 +113,16 @@ proc TclReadLine::clear {} {
 
 proc TclReadLine::clearline {} {
     variable CMDLINE_CURSOR
+    set prompt_len 9
     if {[string match windows $::tcl_platform(platform)]} { 
-        set handle [twapi::get_standard_handle stdout]
-        lassign [ split [ join [ lreplace [ twapi::get_console_screen_buffer_info $handle -cursorpos ] 0 0 ] ] ] col row
-        #col always 9 as length of hammerdb> prompt
-        twapi::fill_console $handle -numlines 1 -position "9,$row"
+        set handle [twapi::get_console_handle stdout]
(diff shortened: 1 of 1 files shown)


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #868, opened 2026-04-19
Improve performance profile comparison,  response time reporting, and CI bug fixes

sm-shaw:
Profile comparison is now clearer, select a Base and a New profile instead of picking two and have HammerDB order. Allows user to pick comparison order of New relative to Base. 
compare
compare2
Response time charts made clearer with interactive boxplot first.
boxpl
Summary page cleanup removed repeated job id/timestamp from chart titles 
Added a shutdown after profile runs to avoid leaving running instances
Removed extra trailing parentheses in system info display, introduced by wapp-subst
system

### DIFF
--- a/modules/jobs-1.0.tm
+++ b/modules/jobs-1.0.tm
@@ -1646,12 +1646,13 @@ proc wapp-page-jobs {} {
             set ::profile_dbdesc [dict create]
         }
         wapp-subst {<h3 class="title">TPROC-C Performance Profiles</h3>}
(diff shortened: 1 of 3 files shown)


jamiere-msft (approved):

sm-shaw:
Summary page for a job now looks as follows  with data followed by charts:
image

abondvt89 (approved):

abondvt89:

PR #877, opened 2026-04-30
Fix form warnings for proxy

sm-shaw:
In the webservice when clicking on Run Benchmark and it is running under nginx it warns that the data it is about to submit is unsafe as it uses the http url instead of https. This PR fixes that issues so no warning is given.


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #878, opened 2026-05-08
Fix profile compare form and hide unsupported from automation

sm-shaw:
Implements a similar fix to #877 for the Jobs page so profile comparisons do not warn about unsecure data by making the page reference relative. 
Also comments out Oracle, SQL Server and Db2 from the initial dropdown for CI/Automation so only MySQL, PostgreSQL and MariaDB are visible.


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #879, opened 2026-05-09
Fix arithmetic in PostgreSQL stock update

sm-shaw:
PR fixes the arithmetic in the PostgreSQL TPROC-C stock update. Tested without impact on performance. 
Also remove stale/misleading comments from Oracle stored procedures as the HammerDB implementation was originally written from the example code rather than the text description in the specification and some of the details diverge between the examples and description. Intention is not to implement any changes that break backward compatibility in performance results.


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #885, opened 2026-05-21
Add database metrics options for MySQL and MariaDB

sm-shaw:
This PR adds database metrics configuration dialogs for MySQL and MariaDB and enables the existing CPU Metrics Only option across database metrics.

Changes included:

- Route MySQL and MariaDB metrics options to database-specific dialogs instead of the generic metrics options.
- Add `metmysqlopts` and `metmariaopts` with the standard connection parameters, SSL options, TPROC-C/TPROC-H user fields, agent controls, and CPU Metrics Only setting.
- Enable the existing CPU Metrics Only option for Oracle and PostgreSQL.
- Fix `cpu_only` handling so it is only defaulted to `false` when unset, rather than always being reset to `false`.

This fixes the issue where MySQL and MariaDB users could start metrics collection without first being directed to configure the required database connection parameters.

It also enables the GUI user to set CPU metrics only and get the CPU display rather than having to configure Oracle, PostgreSQL, MariaDB and MySQL metrics before being able to access the CPU data. This will then be included with a default database report. 

The CLI enable CPU/IO-only by default so is not affected.

### DIFF
--- a/src/generic/gened.tcl
+++ b/src/generic/gened.tcl
@@ -2865,12 +2865,18 @@ proc job_options {} {
 }
 
 proc metricsopts {} {
-    #Database metrics based on Active Session History, currently Oracle and PostgreSQL
+    # Database metrics based on Active Session History:
+    # Oracle, PostgreSQL, MySQL and MariaDB
     global rdbms
+
     if { $rdbms eq "Oracle" } {
         metoraopts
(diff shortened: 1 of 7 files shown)


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #892, opened 2026-06-14
Fix virtual user state across mixed automated, GUI and CLI runs

sm-shaw:
Testing mixed environments from automated web driven workloads then GUI and vice versa allows state to become mixed and require resetting. Not functional errors but improve UX. 
- Enables logtotemp in the CI listener so browser/webhook-dispatched jobs produce output in the web log tail.
- Removes the blocking GUI tk_messageBox when virtual-user logging is enabled, replacing it with console output.
- Adds MySQL and MariaDB to the database metrics-thread exclusion logic, so dbmon threads are not mistaken for active virtual users.
- Clears stale jobs_profile_id during normal GUI Create Virtual Users when autopilot is disabled, preventing a previous profile/autopilot run from being used in a single GUI run which will not be part of profile.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1193,6 +1193,7 @@ Version 6.0 June 2026
 Pull Requests & Issues
 
 Fix virtual user state across mixed automated, GUI and CLI runs #892
+Simplify artifact submit process #891
 Update TPROC-C automated schema sizing #889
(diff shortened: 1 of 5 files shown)


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #893, opened 2026-06-17
Fix AWS cloud metadata detection

sm-shaw:
- Fix AWS IMDSv2 token request by adding an explicit empty PUT body.
- Avoid returning from inside catch blocks in cloud detection.
- Apply the same safe detection pattern to AWS, Azure and GCP.
- Also minor fix print usage text typo from ci||dict to ci|dict.

Previously system discovery would return cloud not detected, after fix and extensive test cloud instance is populated in the system report.

### DIFF
--- a/agent/agent
+++ b/agent/agent
@@ -221,54 +221,73 @@ return $sysinfo
 proc DetectCloudInstance {} {
 #Detect cloud environment by checking metadata services
 #AWS - IMDSv2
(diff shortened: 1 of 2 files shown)


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #894, opened 2026-06-18
Fix SQL Server GUI Scale Factor Option

sm-shaw:
Bug Introduced with PR #835  “Update TPROC-H GUI Options for all Scale Factors”.
Before PR SQL Server TPROC-H used the correct variable:
-variable mssqls_scale_fact
in the scale factor radio buttons.
After the PR merge commit, the same SQL Server dialog had been changed to:
-variable db2_scale_fact
Result is that SQL Server GUI Scale Factor dialog always reverts back to 1.
Fix restores correct behaviour.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1192,6 +1192,7 @@ Version 6.0 June 2026
 
 Pull Requests & Issues
 
+Fix SQL Server GUI Scale Factor Option #894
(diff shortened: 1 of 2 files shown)


jamiere-msft (approved):

abondvt89 (approved):

abondvt89:

PR #862, opened 2026-04-07
Add Azure Citus and Citus Load Balancer support for PostgreSQL

hqakhtar:
This PR adds support for Azure Citus (Azure Elastic Cluster) environments and Citus load balancer port routing for PostgreSQL TPC-C workloads.

### Changes
**Azure Citus support (pg_azure_citus)**
- 
Add pg_azure_citus connection setting to prevent CREATE, ALTER and DROP database operations in Azure Citus managed environments
- Add pg_citus_loadbalancer connection setting for specifying the Citus load balancer port
- Update connection options UI to expose Azure Citus and load balancer port settings
- Add corresponding XML config entries in [`config/postgresql.xml`](https://github.com/TPC-Council/HammerDB/pull/862/changes#diff-adfd032ccd56fec58c14d88cd524a6d9b69b944410c3cd37d6333eee6c8c7e0b)

**Citus load balancer port routing (pg_citus_loadbalancer)**
- Route data population workers through the load balancer port during schema build when Azure Citus and Citus Compatible modes are enabled
- Route OLTP benchmark worker VUs through the load balancer port for non-timed, timed sync, and timed async driver scripts
- DDL operations remain routed through the coordinator via the standard PostgreSQL port
- Add GetDataPort helper to determine the correct port based on configuration

**Documentation**
- Add DocBook documentation entry for the Citus Load Balancer Port setting
- Configuration
- Two new connection settings under [`config/postgresql.xml`](https://github.com/TPC-Council/HammerDB/pull/862/changes#diff-adfd032ccd56fec58c14d88cd524a6d9b69b944410c3cd37d6333eee6c8c7e0b)

### How it works

When `pg_azure_citus` and `pg_cituscompat` are both enabled and `pg_citus_loadbalancer` is set to a non-zero value:
- Schema build: Monitor VU performs DDL on the coordinator (standard port), worker VUs populate data through the load balancer port
- OLTP benchmark: Worker VUs connect through the load balancer port; monitor VU uses the coordinator port for admin tasks (DRITA snapshots, vacuum, transaction counting)

PR #895, opened 2026-06-18
Add Azure Citus and Citus Load Balancer support for PostgreSQL

hqakhtar:
This PR adds support for Azure Citus (Azure Elastic Cluster) environments and Citus load balancer port routing for PostgreSQL TPC-C workloads.

This PR fixes the issue with tcstart.

### Changes
**Azure Citus support (pg_azure_citus)**
- 
Add pg_azure_citus connection setting to prevent CREATE, ALTER and DROP database operations in Azure Citus managed environments
- Add pg_citus_loadbalancer connection setting for specifying the Citus load balancer port
- Update connection options UI to expose Azure Citus and load balancer port settings
- Add corresponding XML config entries in [`config/postgresql.xml`](https://github.com/TPC-Council/HammerDB/pull/862/changes#diff-adfd032ccd56fec58c14d88cd524a6d9b69b944410c3cd37d6333eee6c8c7e0b)

**Citus load balancer port routing (pg_citus_loadbalancer)**
- Route data population workers through the load balancer port during schema build when Azure Citus and Citus Compatible modes are enabled
- Route OLTP benchmark worker VUs through the load balancer port for non-timed, timed sync, and timed async driver scripts
- DDL operations remain routed through the coordinator via the standard PostgreSQL port
- Add GetDataPort helper to determine the correct port based on configuration

**Documentation**
- Add DocBook documentation entry for the Citus Load Balancer Port setting
- Configuration
- Two new connection settings under [`config/postgresql.xml`](https://github.com/TPC-Council/HammerDB/pull/862/changes#diff-adfd032ccd56fec58c14d88cd524a6d9b69b944410c3cd37d6333eee6c8c7e0b)

### How it works

When `pg_azure_citus` and `pg_cituscompat` are both enabled and `pg_citus_loadbalancer` is set to a non-zero value:
- Schema build: Monitor VU performs DDL on the coordinator (standard port), worker VUs populate data through the load balancer port
- OLTP benchmark: Worker VUs connect through the load balancer port; monitor VU uses the coordinator port for admin tasks (DRITA snapshots, vacuum, transaction counting)

### DIFF
--- a/DocBook/docs.xml
+++ b/DocBook/docs.xml
@@ -4197,6 +4197,15 @@ GO
                 </row>
 
                 <row>
+                  <entry>Azure Citus</entry>
+
+                  <entry>Prevents CREATE, ALTER and DROP database operations
+                  for Azure Citus (Azure Elastic Cluster) like managed
+                  environment.
+                  </entry>
+                </row>
+
+                <row>
                   <entry>PostgreSQL Superuser</entry>
 
                   <entry>The PostgreSQL Superuser is a user with sufficient
--- a/DocBook/docs.xml
+++ b/DocBook/docs.xml
@@ -4206,6 +4206,19 @@ GO
                 </row>
 
                 <row>
+                  <entry>Citus Load Balancer Port</entry>
+
+                  <entry>The port of the Citus load balancer used to distribute
+                  connections across worker nodes. This setting is enabled when
+                  Azure Citus is enabled. When set and Citus Compatible mode is
+                  also enabled, schema data population and OLTP benchmark
+                  workload connections will be routed through this port instead
+                  of the standard PostgreSQL port. DDL operations are always
+                  routed through the coordinator via the standard PostgreSQL
+                  port. The default value is 5432.</entry>
+                </row>
+
+                <row>
                   <entry>PostgreSQL Superuser</entry>
 
                   <entry>The PostgreSQL Superuser is a user with sufficient
--- a/config/connectpool/pgcpool.xml
+++ b/config/connectpool/pgcpool.xml
@@ -3,6 +3,7 @@
     <c1>
 	<pg_host>host1</pg_host>
         <pg_port>5432</pg_port>
+	<pg_azure_citus>false</pg_azure_citus>
 	<pg_sslmode>prefer</pg_sslmode>
 	<pg_user>tpcc</pg_user>
         <pg_pass>tpcc</pg_pass>
@@ -11,6 +12,7 @@
     <c2>
 	<pg_host>host2</pg_host>
         <pg_port>5432</pg_port>
+	<pg_azure_citus>false</pg_azure_citus>
 	<pg_sslmode>prefer</pg_sslmode>
 	<pg_user>tpcc</pg_user>
         <pg_pass>tpcc</pg_pass>
@@ -19,6 +21,7 @@
     <c3>
 	<pg_host>host3</pg_host>
         <pg_port>5432</pg_port>
+	<pg_azure_citus>false</pg_azure_citus>
 	<pg_sslmode>prefer</pg_sslmode>
 	<pg_user>tpcc</pg_user>
         <pg_pass>tpcc</pg_pass>
--- a/config/postgresql.xml
+++ b/config/postgresql.xml
@@ -4,6 +4,7 @@
         <pg_host>localhost</pg_host>
         <pg_port>5432</pg_port>
         <pg_sslmode>prefer</pg_sslmode>
+        <pg_azure_citus>false</pg_azure_citus>
     </connection>
 	<tpcc>
         <schema>
--- a/config/postgresql.xml
+++ b/config/postgresql.xml
@@ -5,6 +5,7 @@
         <pg_port>5432</pg_port>
         <pg_sslmode>prefer</pg_sslmode>
         <pg_azure_citus>false</pg_azure_citus>
+        <pg_citus_loadbalancer>5432</pg_citus_loadbalancer>
     </connection>
 	<tpcc>
         <schema>
--- a/scripts/python/postgres/tprocc/pg_tprocc_buildschema.py
+++ b/scripts/python/postgres/tprocc/pg_tprocc_buildschema.py
@@ -9,6 +9,7 @@ diset('connection','pg_host','localhost')
 diset('connection','pg_port','5432')
 diset('connection','pg_sslmode','prefer')
 diset('connection','pg_azure_citus','false')
+diset('connection','pg_citus_loadbalancer','5432')
 
 vu = tclpy.eval('numberOfCPUs')
 warehouse = int(tclpy.eval('numberOfWHs'))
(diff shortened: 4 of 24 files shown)


sm-shaw:
Many thanks for the PR. HammerDB 6.0 is now in the final release cycle, so changes are being limited to release fixes only.
As this PR adds new functionality, we’ll defer review until after 6.0 is released and consider it for the HammerDB 6.1 development cycle.
Many thanks again for the contribution.

sm-shaw (changes requested):
Many thanks for coming back with this, and apologies that review has taken longer than it should due to the HammerDB v6.0 release.  We do want to support Citus, including Azure Elastic Cluster, in HammerDB.

I have reviewed the implementation in detail and there are four functional issues that currently block this PR, similar to #862. 

1. Schema checking is broken for both TPROC-C and TPROC-H. The generated schema-check calls now pass an extra Azure-specific value, but the corresponding procedures still expect the original arguments, so the checks fail with a wrong number of arguments error.
2. Cluster Connect Pool is broken. The connection setup now includes an extra Azure-specific value, but the existing ConnectToPostgres procedure has not changed and still expects the normal PostgreSQL connection arguments.
3. The asynchronous TPROC-C path is incomplete. It uses citus_compatible and citus_lb_port, but those variables are not defined in the generated asynchronous script.
4. TPROC-H schema creation can fail in Azure mode. The code reserves a database SQL command but, when Azure mode skips that database operation, no command is actually created. It later attempts to execute the missing command, resulting in an undefined sql(...) variable error.

I think there could be a much simpler approach.

HammerDB already has pg_cituscompat for TPROC-C. When that is enabled, HammerDB already knows that it is running a Citus-compatible TPROC-C workload. From the existing coordinator connection, it should then determine whether that Citus environment is Azure Elastic Cluster and apply only the Azure-specific behaviour that is actually required.
If we need a new parameter for TPROC-H lets introduce pg_tpch_cituscompat to TPROC-H as well for compatibility or do a more focused PR for TPROC-C first - we get that included and then do TPROC-H. 

This means we should not introduce another Azure-specific configuration parameter, and we should not add Azure-specific settings to the standard Tcl/Python scripts or existing PostgreSQL connection paths.

The existing PostgreSQL implementation should remain unchanged wherever possible. Coordinator, transaction counter, metrics, schema checking and administrative connections should continue to work as they currently do.

So ideally, only the Citus-specific parts should change: when pg_cituscompat is enabled, determine whether the connected Citus environment is Azure Elastic Cluster; When a pre-provisioned database exists, use that database rather than attempting database creation or deletion; use a Citus load-balancer port for schema population workers and benchmark VUs where required. ConnectToPostgres should therefore remain unchanged. If worker connections need to use a different port, that should be determined before making the connection rather than changing the existing PostgreSQL connection interface.

This should come back as a much smaller change centred on the existing pg_cituscompat support.  The branch will also need rebasing onto current master.

hqakhtar:
Thank you for the review. Let me review the changes and resolve these issues.

hqakhtar:
Updated todo list

Thank you for the detailed review @sm-shaw. I propose revising the PR as follows:

- Keep Azure Elastic Cluster selection explicit because there is no reliable SQL property that definitively identifies the Azure service.
- Rename the settings to:
  - `pg_azure_elastic_cluster`, Boolean, default `false` - Require `pg_cituscompat` to be set 
  - `pg_citus_loadbalancer`, optional port, default `0` - Require `pg_cituscompat` to be set
- Keep `ConnectToPostgres` unchanged. Azure-specific values will not be added to generic connection arguments or connection-pool definitions.
- Use `pg_port` for coordinator, DDL, schema checks, metrics, transaction counting, and administrative operations. This remains valid when the configured port is not `5432`.
- Use `pg_citus_loadbalancer` only for schema population workers and benchmark VUs. When it is `0`, workers also use `pg_port`.

This addresses the four functional issues by:

1. Restoring the original TPROC-C and TPROC-H schema-check signatures and calls.
2. Removing the extra Azure argument from Cluster Connect Pool.
3. Passing a resolved worker port to all TPROC-C driver modes, including asynchronous, so they do not reference undefined variables.
4. Removing TPROC-H Azure changes from this PR and handling TPROC-H later through a focused Citus-compatible change.

In Azure Elastic Cluster mode, HammerDB will require a pre-existing database and skip `CREATE DATABASE`, `ALTER DATABASE`, and `DROP DATABASE`. I will also rebase onto current `master` and test standard PostgreSQL, Citus, Connect Pool, and all TPROC-C driver modes.

sm-shaw:
Many Thanks, this proposed approach looks much closer to what we need.

Keeping ConnectToPostgres and the existing PostgreSQL coordinator, administrative, schema-check, metrics and transaction-counter paths unchanged addresses the main concern with the previous implementations.

I am also happy with using an explicit pg_azure_elastic_cluster setting if there is no reliable way to identify the Azure service from PostgreSQL itself, provided it only has an effect when pg_cituscompat is enabled. Using 0 to disable pg_citus_loadbalancer by default also makes sense.

Keeping this PR focused on TPROC-C and leaving TPROC-H for a separate change is the right approach.

Please go ahead with this design. The key point is that the resulting change should be much smaller and centred on the existing Citus-compatible path, with the standard PostgreSQL behaviour remaining unchanged. The proposed regression testing of standard PostgreSQL, Citus, Connect Pool and all TPROC-C driver modes should then give us what we need to review it for it to be included in the HammerDB v6.1 release.

sm-shaw:
Hi @hqakhtar,

Just checking on timing for this one.

We are currently targeting HammerDB 6.1 for around e/o Oct. To have a realistic chance of including this in 6.1, we would ideally need the revised implementation rebased, tested and ready for review by the end of September.

Do you think that timeframe is realistic?

If not, no problem, but we would probably need to move it to the next development cycle rather than hold the 6.1 release open for it, so wanted to make you aware.

Many thanks.

hqakhtar:
> Hi @hqakhtar,
> 
> Just checking on timing for this one.
> 
> We are currently targeting HammerDB 6.1 for around e/o Oct. To have a realistic chance of including this in 6.1, we would ideally need the revised implementation rebased, tested and ready for review by the end of September.
> 
> Do you think that timeframe is realistic?
> 
> If not, no problem, but we would probably need to move it to the next development cycle rather than hold the 6.1 release open for it, so wanted to make you aware.
> 
> Many thanks.

Testing my changes at the moment. I will hopefully be able to share the PR this week with all concerns addressed.

hqakhtar:
> Hi @hqakhtar,
> 
> Just checking on timing for this one.
> 
> We are currently targeting HammerDB 6.1 for around e/o Oct. To have a realistic chance of including this in 6.1, we would ideally need the revised implementation rebased, tested and ready for review by the end of September.
> 
> Do you think that timeframe is realistic?
> 
> If not, no problem, but we would probably need to move it to the next development cycle rather than hold the 6.1 release open for it, so wanted to make you aware.
> 
> Many thanks.

I've updated the PR @sm-shaw . The update adds:

- pg_citus_azure_elastic_cluster
- pg_citus_loadbalancer (0 uses pg_port)
- LB routing applies only when both Citus and Azure modes are enabled. pg_port remains used for coordinator/admin/check/metrics traffic.

Verified:

- Schema-check API unchanged.
- ConnectToPostgres remains six arguments.
- Sync, async, and Connect Pool routing pass.
- LB messages appear only when routing is active.
- Standard PostgreSQL, plain Citus, fallback, and TPROC-H remain unchanged.

PR #896, opened 2026-06-18
Enable WAL and normal synchronous mode for the jobs SQLite DB

hqakhtar:
# Add WAL and synchronous=NORMAL for the jobs SQLite DB

This PR updates the jobs SQLite backend to use `WAL` and `synchronous=NORMAL` for on-disk databases. The change is applied when the jobs database is opened and reopened, while keeping `:memory:` behavior unchanged.

## Changes
- Enable `PRAGMA journal_mode=WAL` for the jobs SQLite DB when it is not `:memory:`
- Set `PRAGMA synchronous=NORMAL` for the same on-disk databases
- Forward worker log lines to the master thread asynchronously

## Why
- WAL allows readers and writers to operate with less contention on the jobs database
- `synchronous=NORMAL` reduces fsync pressure and improves write throughput for this workload
- Asynchronous log forwarding avoids blocking worker threads on master-thread log handling

## Result
In testing, this improved data loading by around 20% with 100 workers and 400 warehouses. Higher worker counts should benefit even more due to reduced contention on the jobs database and logging path.

### DIFF
--- a/modules/jobs-1.0.tm
+++ b/modules/jobs-1.0.tm
@@ -90,6 +90,10 @@ namespace eval jobs {
     } else {
       catch {hdbjobs timeout 30000}
       #hdbjobs eval {PRAGMA foreign_keys=ON}
+      if { $sqlite_db ne ":memory:" } {
+        catch {hdbjobs eval {PRAGMA journal_mode=WAL}}
+        catch {hdbjobs eval {PRAGMA synchronous=NORMAL}}
+      }
       if { $sqlite_db eq ":memory:" } {
         catch {hdbjobs eval {DROP TABLE JOBMAIN}}
         catch {hdbjobs eval {DROP TABLE JOBTIMING}}
@@ -249,6 +253,10 @@ namespace eval jobs {
     }
     if [catch {sqlite3 hdbjobs $sqlite_db} message ] {} else {
       catch {hdbjobs timeout 30000}
+      if { $sqlite_db ne ":memory:" } {
+        catch {hdbjobs eval {PRAGMA journal_mode=WAL}}
+        catch {hdbjobs eval {PRAGMA synchronous=NORMAL}}
+      }
       if [catch {set tblname [ hdbjobs eval {SELECT name FROM sqlite_master WHERE type='table' AND name='JOBMAIN'}]} message ] {
         puts "Error querying  JOBOUTPUT table in SQLite on-disk database : $message"
         return
--- a/src/generic/genvu.tcl
+++ b/src/generic/genvu.tcl
@@ -305,7 +305,7 @@ proc load_virtual {}  {
                         if { $masterthread eq -1 } { ; } else {
                             lappend Stack $s$nl
                             if { [ llength $Stack ]  > 5 } { set Stack [lreplace $Stack 0 0 ] }
-                            eval [subst {thread::send $masterthread {::Log [ list $id $Stack $s$nl ]}}]
+                            eval [subst {thread::send -async $masterthread {::Log [ list $id $Stack $s$nl ]}}]
                             if { $optlog eq 1 } { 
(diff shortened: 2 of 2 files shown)


sm-shaw:
Many thanks for the PR. This looks like a useful improvement, especially for reducing contention in the jobs SQLite database and improving higher-VU load performance.

HammerDB 6.0 is now in the final release cycle, so changes are being limited to release fixes only before the v6.0 tag. As this changes the jobs database behaviour and the log forwarding path, we do not want to risk introducing a regression at this stage.

We’ll keep this open and review/test it properly for the HammerDB 6.1 development cycle.

Many thanks again for the contribution.

sm-shaw:
For commit https://github.com/TPC-Council/HammerDB/pull/896/changes/77173b8a38b0c9886713860682205db814e9d1e5 Log was synchronous by design because screen updates are inherently serialised by the GUI. On both Windows and Linux, an application's window updates are handled by a single GUI/display thread rather than being updated concurrently by multiple worker threads. Making the VU sends asynchronous therefore does not make display processing parallel; it just allows more updates to queue waiting for the same single consumer. This is why Log is synchronous while the separate logtofile path, which has no display dependency, is already asynchronous.  This does not mean that asynch cannot be used for the CLI but it cannot be set globally for both - when the GUI is used we want all updates to the master thread synchronous. I will look into distinguishing between the two. The SQLite settings should be OK as we have already tested WAL mode with sqlite3-rsync to replicate HammerDB changes.

sm-shaw:
To confirm that for me the 2nd change making Log async causes the GUI to lock up with a large number of messages. With sync it can slow down but doesn't lock and the display can be closed.  Testing with async and `thread::configure -eventmark` does not rectify the issue. This confirms the original design why Log was sync and Logtofile was async.
image

sm-shaw:
On the SQLite change specifically, we saw reproducible abnormal behaviour after enabling journal_mode=WAL with synchronous=NORMAL.

Schema build/delete operations completed and the corresponding rows were still present in JOBMAIN, but the CLI began returning:

Schema Build (No jobid)
Schema Delete (No jobid)

In the same running CLI session, the jobs command could then report that no jobs were present at all. Restarting HammerDB made the existing jobs visible again. We also found that reverting the database setting alone did not restore normal behaviour; clearing the temporary Jobs state completely was required.

Separately, HammerDB currently ships SQLite 3.49.1. SQLite has documented a WAL-reset corruption bug affecting versions from 3.7.0 through 3.51.2 https://www.sqlite.org/wal.html#the_wal_reset_bug. The bug is a concurrency race involving multiple connections in separate threads or processes accessing the same WAL database while writes and checkpoints overlap. Under the specific race condition, WAL state can become inconsistent and a later checkpoint can fail to copy committed pages into the main database correctly. The issue was fixed in SQLite 3.51.3, with fixes also backported to some maintained branches. Specifically, HammerDB v6.0 currently uses Tcl 9.0.2 with SQLite 3.49.1, which falls within the SQLite versions affected by the documented WAL-reset concurrency bug. The current Tcl 9.1 development release (9.1b0) bundles SQLite 3.53.0,

We are not saying that the behaviour observed in HammerDB is necessarily this SQLite bug. However, the proposed change enables WAL while HammerDB is using an SQLite release known to contain a WAL concurrency issue, and our own testing independently produced serious and reproducible Jobs-state problems after making that change.

Given HammerDB's multithreaded architecture, the importance of the Jobs database, and the stability of the existing SQLite configuration, we do not think the potential performance benefit justifies introducing WAL here. We will therefore retain the existing SQLite configuration but can revisit this changed when HammerDB moves to a Tcl releases with SQLite 3.53.0

sm-shaw:
Thanks again for the contribution. Based on the testing above, we’re going to retain the existing implementation and close this PR, however may be able to revisit the SQlite change from SQLite 3.53.0.

PR #897, opened 2026-06-19
Fix NULL Oracle Version reported for TPROC-H

sm-shaw:
The query used to select the Oracle version number for TPROC-C does not work for TPROC-H as the user does not have the correct privileges. Changed SQL for a query that works for a user without system privileges. TPROC-C can remain unmodified as it is system user running the query.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1192,6 +1192,7 @@ Version 6.0 June 2026
 
 Pull Requests & Issues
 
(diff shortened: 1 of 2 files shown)


jamiere-msft (approved):

sm-shaw:
As single-line bug fix, merging after 2 approvals to enable HammerDB v6.0 release preparation before tagging the release.

PR #904, opened 2026-08-03
Fix pg_vacuum end-of-run ordering and false VU-shutdown-timeout report

### LINKED ISSUE #903: pg_vacuum: end-of-run CHECKPOINT runs before VACUUM (wrong order), and a slow VACUUM can trigger a false "Virtual Users remain running" error
**Guidance**

**Describe the bug**
1.  end-of-run CHECKPOINT runs before VACUUM (wrong order)
When pg_vacuum is enabled, the code will do first a checkpoint and then a vacuum. However, the VACUUM phase will generate dirty buffers (due to visibility map updates/dead tuples removal) and a CHECKPOINT issued before doesn't flush the dirty pages vacuum generated.

2.  slow VACUUM can trigger a false "Virtual Users remain running" error
Within large datasets, I've seen misleading errors like above and digging further i noticed that the root cause for it is that vacuum/checkpoint (pg_vacuum) execute as blocking pg_exec calls inside the monitor VU's own thread. In some cases (large tables), vacuum doesn't complete in the "rampup + duration + keepalive_margin" window and errors out. A workaround would be to increase drastically the keepalive_margin parameter but that would just be a guess about how long vacuum actually takes.

**To Reproduce**
1. Generate 25k wh dataset
2. run a scaling test of like 8-16-32 vus
3. enable pg_vacuum
4. See error at the end of the 1st run 

**Expected behavior**
Vacuum + check point should happen in this order and HammerDB should wait for vacuum to complete before invoking vudestroy

 - Version: 6.0
 - Build: [release 6.0]

 - UI: [CLI]

 - Server OS: [RHEL9.8]
 - Client OS: [RHEL9.8]

**Additional context**

alfer-edb:
See: https://github.com/TPC-Council/HammerDB/issues/903

### DIFF
--- a/src/generic/gencli.tcl
+++ b/src/generic/gencli.tcl
@@ -1809,14 +1809,17 @@ proc _runtimer { seconds } {
     upvar timevar timevar
     proc runtimer_loop { seconds } {
         upvar elapsed elapsed
-        incr elapsed
         upvar timevar timevar
         set rcomplete [vucomplete]
+        set vacuum_active [expr {[tsv::exists application vacuum_running] && [tsv::get application vacuum_running]}]
+        if { !$vacuum_active } {
+            incr elapsed
+        }
         if { ![ expr {$elapsed % 60} ] } {
             set y [ expr $elapsed / 60 ]
             #putscli "Timer: $y minutes elapsed"
         }
-        if {!$rcomplete && $elapsed < $seconds } {
+        if {!$rcomplete && ($vacuum_active || $elapsed < $seconds) } {
             ;#Neither vucomplete or time reached, reschedule loop
         catch {after 1000 runtimer_loop $seconds }} else {
             #putscli "keepalive returned after $elapsed seconds"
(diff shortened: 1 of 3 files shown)


sm-shaw (changes requested):
I have tested and all works fine. I would like to to request a change that as the CLI proc _runtimer is generic the change in gencli.tcl be made specific to PostgreSQL. e.g. I run multiple tests against different databases back to back so it would be good to ensure that vacuum_active could never be live for another database. Also, although it should be clear to most that vacuum means PostgreSQL this is not explicit in the code.  As an example we could have 2 runtimer_loops in runtimer determined on the setting of $rdbms being PostgreSQL.

sm-shaw:
Can we please also add the same change in load_virtual and run_virtual so it is clear that this is for PostgreSQL only and the setting only made for PostgreSQL.

alfer-edb:
Hey @sm-shaw, I made the changes you asked for, have a look at your convenience !

sm-shaw (approved):
Tested the updated changes with PostgreSQL end-of-run VACUUM and CHECKPOINT processing and with a subsequent non-PostgreSQL workload in the same HammerDB session. VACUUM now runs before CHECKPOINT.

abondvt89 (approved):

abondvt89:

PR #911, opened 2026-08-24
Add Quack driver for sqlxtc (TPROC-C)

gburd:
## Summary

This PR adds HammerDB support for **sqlxtc**, a compact networked,
threaded SQL engine that ships as the `06_sqlxtc` example of the
[libxtc](https://codeberg.org/gregburd/libxtc) project. sqlxtc speaks a
deliberately tiny **newline-delimited-JSON wire protocol called "Quack"**
over plain TCP (default port `15432`): the server sends a `{"hello":...}`
banner, the client sends one JSON object per line
(`{"q":"SQL"}`, `{"q":"SQL","params":[...]}`, `{"ping":1}`, `{"quit":1}`),
and the server streams back `{"cols":[...]}` / `{"row":[...]}` /
`{"done":N}` / `{"err":...}`.

A new database is registered as **`quack`** (`dbset db quack`).

## What's in the PR

* **`modules/quacktcl-1.0.tm`** - a dependency-light, **pure-Tcl** Quack
  client (Tcl's built-in `socket` plus a tiny flat-JSON encoder/decoder
  written for the Quack framing). It pulls in no external Tcl package,
  so it runs under the bundled runtime and a bare `tclsh` alike. Command
  surface mirrors the other DB access packages: `quackconnect`,
  `quackexec`, `quacksel` (`-list` / `-flatlist`), `quackcommit`,
  `quackrollback`, `quackclose`, `quackping`, `quackescape`.
* **`src/quack/quackoltp.tcl`** - TPROC-C schema build (`build_quacktpcc`)
  and driver (`loadquacktpcc` / `loadtimedquacktpcc`), adapted from the
  MySQL "no stored procedures" client-SQL driver (the closest existing
  fit).
* **`src/quack/{quackopt,quackotc,quackmet,quackci,quackolap}.tcl`** -
  GUI config dialogs, the transaction-counter thread, metrics/CI stubs,
  and the TPROC-H placeholder.
* **`config/quack.xml`** + registration in **`config/database.xml`** and
  the three `geninit*.tcl` DB-reset lists.
* **`src/quack/README.md`** - full scope and limitation notes.

The connection target is only **host + port** - sqlxtc has no users,
passwords, databases or tablespaces.

## sqlxtc SQL surface (verified against the live server)

**Supported and used:** `CREATE TABLE` (composite `PRIMARY KEY`),
`INSERT` (incl. multi-row `VALUES`), `UPDATE`/`DELETE`,
`BEGIN`/`COMMIT`/`ROLLBACK`, single-table aggregation + `GROUP BY`,
projection `JOIN ... ON` with table-qualified columns, `IN (subquery)`,
`ORDER BY` / `LIMIT` / `BETWEEN`.

**Worked around in the driver:**

* no comma-joins - the driver uses explicit `JOIN ... ON`;
* join projections must be **table-qualified** (`SELECT a.x, b.y`);
* no `SELECT ... FOR UPDATE` and no `START TRANSACTION` - the driver
  uses `BEGIN`;
* the stock-level transaction is expressed as a single-table
  `COUNT(DISTINCT ...)` with an `IN` subquery, because sqlxtc does not
  support `JOIN` combined with aggregation.

**Not available (honest gap list):**

* **`CREATE INDEX`** - sqlxtc has no secondary-index DDL, so the driver
  omits indexes. The schema is correct and queries return correct
  results, but every non-primary-key lookup is a table scan. This is the
  dominant performance limit.
* **`JOIN` + aggregation** (join + `SUM`/`GROUP BY`) returns
  `unsupported SQL (native engine)`.
* No server-side commit statistics or catalog/performance views, so
  **NOPM** is derived from `sum(d_next_o_id)` (as the other drivers do)
  and **TPM** is counted client-side in a shared tsv counter.

## TPROC-C vs TPROC-H

TPROC-C works. **TPROC-H is intentionally not offered for Quack**:
almost every TPC-H query is a multi-table join *with* aggregation (the
unsupported combination above), and several reference queries also need
server-side functions sqlxtc lacks (`pgmedian`, `pgpercentile_cont`,
`array_agg`). `quackolap.tcl` reports this rather than failing partway
through the query set. When sqlxtc gains JOIN-with-aggregation, the
single-table-friendly subset of the analytic queries (e.g. Q1, and the
`IN`-subquery queries, which already run) can be ported.

## How it was exercised

Run end to end against the libxtc `06_sqlxtc` example server (durable
file on local NVMe):

* A **1-warehouse TPROC-C schema builds cleanly and durably**, with the
  correct TPC-C cardinalities: 1 warehouse, 10 districts, 30000
  customers, 100000 items, 100000 stock, 30000 orders, 300557
  order_lines, 9000 new_orders, 30000 history rows.
* The **full transaction mix runs clean**: 298/300 transactions
  committed with **zero SQL errors** (the 2 non-commits are the
  spec-mandated ~1% New-Order rollbacks).
* Throughput is low (order of ~15 NOPM single-connection) and does not
  scale with concurrency on a single warehouse - a direct consequence
  of the missing secondary indexes: each customer/stock/order lookup is
  a full table scan, and concurrent workers contend on the single
  warehouse/district row. This is expected to improve substantially,
  with no driver change, once sqlxtc gains `CREATE INDEX`.

All new Tcl sources pass `info complete`, and the pure-Tcl `quacktcl`
package and the driver's schema-build and transaction procs were run
directly against a live sqlxtc server to validate every statement the
driver emits.


sm-shaw:
Many thanks for the pull request,  the driver implementation looks really interesting, but I don't think sqlxtc is quite ready for inclusion in HammerDB yet. 
A similar discussion is here - https://github.com/TPC-Council/HammerDB/discussions/915, you do provide the driver which is a key step. 
However the current 15 NOPM result and lack of scaling, means it cannot yet provide a comparative TPROC-C performance workload. If you look at the community submitted results coming online https://tpc-council.github.io/hammerdb-results/index.html then this gives an indication of the sort of area of expected performance.
As noted in the discussion earlier we did support Redis, However support was deprecated as performance was low as Redis is single threaded and not a fair representation of its use, compared to what traditional RDBMS could. We also don't test SQLite for a similar reason even though SQLite is a database that HammerDB itself is built on. This sort of high performance enterprise RDBMS test is not the area they are targeted. 
So, overall the bar is quick high for inclusion and at the moment we only directly support the mainstream Cloud RDBMS of Oracle, SQL Server, Db2, PostgreSQL, MySQL and MariaDB. If the scenario for sqlxtc changes we would be happy to revisit but for now the 15 NOPM would not a reasonable user comparison.

sm-shaw:
Thanks again for the contribution and for the work that has gone into implementing and testing the driver.

For now I’m going to close this PR, as the current sqlxtc performance and lack of scaling mean it isn’t yet suitable for comparative TPROC-C testing in HammerDB.

If sqlxtc develops further, particularly with secondary index support and improved performance at higher concurrency, we would be very happy to revisit this.

PR #913, opened 2026-08-25
Fix Linux I/O metric partition double counting

### LINKED ISSUE #909: Title: Linux I/O metrics double count whole disks and partitions
**Guidance**

**Describe the bug**
Linux I/O metrics can double count IOPS and MB/s because HammerDB sums both whole block devices and their partitions from `/proc/diskstats`.

For example, on an AWS EC2 instance with NVMe-backed EBS storage, `/proc/diskstats` can contain both:

```text
nvme0n1
nvme0n1p1
```

The current Linux collector includes both entries. This can cause the same underlying I/O to be counted at both the whole-device and partition level.

The `/proc/diskstats` approach itself is valid. If HammerDB uses only the top-level block device, for example `nvme0n1`, the IOPS and MB/s counters are appropriate. The problem is summing the top-level device and its partitions together.

This was observed on an AWS gp3 EBS volume provisioned for 3,000 IOPS, where HammerDB reported peaks of approximately 6,000 IOPS.

**To Reproduce**
1. Run HammerDB on Linux where `/proc/diskstats` contains both a whole block device and one or more partitions, for example `nvme0n1` and `nvme0n1p1`.
2. Enable HammerDB metrics.
3. Run a workload generating sustained storage I/O.
4. Compare the HammerDB IOPS/MB/s values with the I/O capability of the underlying block device.
5. Inspect `/proc/diskstats` and observe that HammerDB is summing counters for both the whole device and its partitions.

**Expected behavior**
HammerDB should count each underlying block device once.

For Linux, IOPS and MB/s should be calculated from the whole-device `/proc/diskstats` counters only, excluding partition entries.

For example, `nvme0n1` should be counted, while `nvme0n1p1` should not also be added to the total.

The implementation should work correctly for NVMe and SATA/SCSI devices and should also account correctly for device-mapper/LVM configurations.

**Screenshots**
Screenshot showing approximately 6,000 IOPS reported by HammerDB on an AWS gp3 volume provisioned for 3,000 IOPS.

 - Version: 6.0
 - Build: Release download

 - UI: Web Service / Pipelines metrics

 - Server OS: Ubuntu 24.04
 - Client OS: Ubuntu 24.04

**Database Server (please complete the following information):**
 - Database name: MariaDB
 - Database Release Version: Built by HammerDB Pipeline

**Database Client (please complete the following information):**
 - Database client name: MariaDB client library

**Additional context**
Linux I/O metrics were introduced in #880.

The current device filtering is:

```tcl
set dev [lindex $fields 2]
if {[regexp {^(loop|ram|fd|sr|dm-|md)} $dev]} { continue }

set reads [expr {$reads + [lindex $fields 3]}]
set writes [expr {$writes + [lindex $fields 7]}]
set sectors [expr {$sectors + [lindex $fields 5] + [lindex $fields 9]}]
```

This filters some device types by name but does not distinguish whole block devices from partitions.

Windows is not affected by this specific issue because the Windows implementation uses PDH `disk_transfer_rate` and `disk_bytes_rate` counters rather than `/proc/diskstats`.

Image

sm-shaw:
Fixes Linux double counting I/O by skipping partitions and only recording the underlying device.


abondvt89 (approved):

abondvt89:

PR #115, opened 2020-03-30
Add partitioning hints in PostgreSQL neword transaction

marcocitus:
#92 rewrote two of the stored procedure implementations in `pgoltp.tcl` for higher performance. In the neword procedure, [one of the update queries](https://github.com/TPC-Council/HammerDB/pull/92/files#diff-d897f78b12b2f4c500c61976159deb04R547) now joins the stock table with the unnest function on the s_w_id column. While the PostgreSQL planner is able to use the primary key index during a join, it is not able to prune away partitions when the `stock` table is partitioned based on the join. Hence, the neword procedure now always updates all partitions of the stock table even if it is partitioned on s_w_id and there is only a single value of s_w_id.

We observed a significant drop in performance when running the new stored procedure implementation against a PostgreSQL cluster that uses the [Citus extension](https://github.com/citusdata/citus) to hash partition the data across multiple PostgreSQL servers, which was caused by the neword transaction spanning across all servers. A similar problem occurs when using PostgreSQL's built-in partitioning mechanism, which can be used to improve performance by reducing fragmentation and can also used as a rudimentary clustering solution in combination with postgres_fdw.

This PR addresses the issue by adding a `stock.s_w_id = ANY (supply_wid_array)` filter to the update(s) in the neword procedure to enable partition pruning by explicitly restricting the value of s_w_id to any of the values in the supply_wid_array. Adding the filter does not affect the semantics of the query, since this restriction is already implicitly present through the join with the unnested array. The performance penalty of the extra filter is negligible given that the array is always small (at most 15 elements), and the filter can result in better query plans even for regular tables.

To demonstrate, the following queries show how the query plan for the update in question changes before and after applying the changes in this PR when using PostgreSQL's built-in hash-partitioning ([example](https://gist.github.com/marcocitus/e9de3d71da2ec411c47e2dcb9e461b5b)).

```sql
PREPARE su_before(INT[], SMALLINT[], SMALLINT[], NUMERIC(5,2)[], NUMERIC, NUMERIC, NUMERIC) AS
                WITH stock_update AS (
                UPDATE stock
               SET s_quantity = ( CASE WHEN s_quantity < (item_stock.quantity + 10) THEN s_quantity + 91 ELSE s_quantity END) - item_stock.quantity
                          FROM UNNEST($1, $2, $3, $4)
                                   AS item_stock (item_id, supply_wid, quantity, price)
                         WHERE stock.s_i_id = item_stock.item_id
                           AND stock.s_w_id = item_stock.supply_wid
[...]

PR #920, opened 2026-09-11
Add VillageSQL database engine (TPROC-C / TPROC-H)

tomas-villagesql:
## Summary

Adds VillageSQL as a first-class database engine for TPROC-C and TPROC-H, following the existing data-driven engine pattern (config/database.xml `<villagesql>` block, `src/villagesql/` sources, generic-layer wiring, and `scripts/tcl/villagesql/` benchmark drivers). VillageSQL is MySQL wire-compatible, so the engine uses the `mysqltcl` driver.

## Layout

Follows the established engine convention of a descriptive key and a short prefix:

- **key** `villagesql` -> directory `src/villagesql/`, config `<villagesql>`, `config/villagesql.xml`
- **prefix** `vsql` -> filenames `vsql*.tcl`, `dbset db vsql`, options `vsql_*`
- **display name** VillageSQL

## Files

- `src/villagesql/vsql{oltp,olap,opt,otc,ci,met}.tcl` -- engine sources. `vsqlmet.tcl` is a thin shim that reuses the identical MySQL metrics implementation rather than duplicating ~3000 lines.
- `config/villagesql.xml`, `config/database.xml` (`<villagesql>`), `config/connectpool/vsqlcpool.xml`
- `scripts/tcl/villagesql/{tprocc,tproch}/` -- standalone benchmark drivers (build/run/check/delete + shell wrappers), matching the per-engine `scripts/tcl/<engine>/` trees.
- Generic-layer wiring: `geninit*`, `gengen`, `gened`, `genmetrics`, `genvu`; modules `jobs`, `pipelines`, `tpcccommon`

## Testing

Validated end-to-end via a from-source runner (tclsh + mysqltcl) against a live VillageSQL server:

- TPROC-C: build -> run -> check, all 4 consistency checks pass
- TPROC-C multi-VU (4 warehouses, 4 VUs)
- TPROC-H: schema build + all 22 queries

## Scope

Engine only. The CI provider integration (a `<VillageSQL>` block in `config/ci.xml` that builds and benchmarks the server) is a separate patch.


sm-shaw (changes requested):
The PR builds and we have now tested it against a current VillageSQL server. Before approval, I think the following points should be addressed.

### 1. VillageSQL Metrics currently use the MySQL configuration rather than the VillageSQL configuration

`vsqlmet.tcl` avoids duplicating the MySQL metrics implementation by aliasing `vsqlmetrics` directly to `mysqlmet::mysqlmetrics`.

Reusing the common MySQL `performance_schema` metrics implementation makes sense, but the MySQL metrics implementation explicitly reads `configmysql`, uses the MySQL SSL setup and references the `mysql_*` connection variables.

This means the VillageSQL Metrics dialog can contain the correct `vsql_*` values, but starting Metrics then enters code which reads the MySQL configuration.

That is a functional bug, and the testing described in the PR does not appear to cover Metrics.

The common metrics implementation can still be shared, but the connection and configuration handling needs to use `configvillagesql` and the corresponding `vsql_*` settings when VillageSQL is selected.

### 2. OceanBase has been copied into the VillageSQL implementation

The VillageSQL implementation currently contains OceanBase-specific configuration and UI, including OceanBase compatibility, tenant name and partition settings, together with OceanBase-specific handling in TPROC-H.

We do not want this carried forward into VillageSQL.

VillageSQL is becoming a database category in its own right. OceanBase is separately becoming its own database category. VillageSQL should therefore contain VillageSQL support rather than inherit the historical OceanBase compatibility functionality from MySQL.

This is a good example of why separate VillageSQL support is the right approach. VillageSQL, MySQL and OceanBase can evolve independently rather than carrying unrelated compatibility options through another database implementation.

### 3. Pipelines is exposed before VillageSQL CI exists

PR #920 already adds VillageSQL to the Pipelines database mappings and UI, so VillageSQL is presented as a first-class database that can be selected for automated testing.

The current VillageSQL Pipelines interface already exposes:

```text
VillageSQL

Ref:
  tag, branch or commit SHA

TPROC-C:
  Single
  Profile
  Compare
  Full Durability + I/O intensive

TPROC-H:
  Single
```

**Screenshot 1: VillageSQL CI Pipelines UI**

Image

This is exactly the direction we want to take the integration. VillageSQL should reuse the MySQL foundation where appropriate, but have its own first-class identity in HammerDB.

However, `vsqlci.tcl` explicitly says that it is only a stub and that the clone/build/install/test harness has not been implemented.

That means a significant part of the functionality already presented by the Pipelines interface is missing underneath.

HammerDB CI is much more than simply having a database in a dropdown. For the existing open source databases, HammerDB can automatically create and validate the environment required to clone upstream source, build it, install it, start it and execute HammerDB tests against the resulting server.

It can be triggered automatically through Pipelines, and the same process can also be run directly by anyone using HammerDB CLI.

For example:

```text
hammerdb>cifix
Running cifix...
CI FIX: starting
CI FIX: provider MariaDB
CI FIX: provider PostgreSQL
CI FIX: provider MySQL
CI FIX: directory exists /opt/hammerdb-ci
CI FIX: directory exists /opt/hammerdb-ci/TMP
CI FIX: directory exists /opt/hammerdb-ci/providers/mariadb
CI FIX: directory exists /opt/hammerdb-ci/providers/mariadb/builds
CI FIX: directory exists /opt/hammerdb-ci/providers/mariadb/config
CI FIX: directory exists /opt/hammerdb-ci/providers/mariadb/installs
CI FIX: directory exists /opt/hammerdb-ci/providers/mysql
CI FIX: directory exists /opt/hammerdb-ci/providers/mysql/builds
CI FIX: directory exists /opt/hammerdb-ci/providers/mysql/config
CI FIX: directory exists /opt/hammerdb-ci/providers/mysql/installs
CI FIX: directory exists /opt/hammerdb-ci/providers/postgresql
CI FIX: directory exists /opt/hammerdb-ci/providers/postgresql/builds
CI FIX: directory exists /opt/hammerdb-ci/providers/postgresql/config
CI FIX: directory exists /opt/hammerdb-ci/providers/postgresql/installs
CI FIX: file exists /opt/hammerdb-ci/providers/mariadb/config/maria.cnf
CI FIX: file exists /opt/hammerdb-ci/providers/mariadb/config/mariaio.cnf
CI FIX: file exists /opt/hammerdb-ci/providers/mysql/config/my.cnf
CI FIX: file exists /opt/hammerdb-ci/providers/mysql/config/myio.cnf
CI FIX: file exists /opt/hammerdb-ci/providers/postgresql/config/postgresql.conf
CI FIX: file exists /opt/hammerdb-ci/providers/postgresql/config/postgresqlio.conf
CI FIX: validating
CI VALIDATE: provider MariaDB
CI VALIDATE: provider PostgreSQL
CI VALIDATE: provider MySQL
CI VALIDATE: PASSED
CI FIX: COMPLETE
```

VillageSQL is completely absent from that workflow today.

Because VillageSQL is open source, it is particularly well suited to this capability. HammerDB should be able to take the tag, branch or commit SHA entered in the Pipelines interface, automatically clone and build VillageSQL from that source, install it, start the resulting server and execute the selected benchmark against it.

That gives VillageSQL automated regression testing and also allows anyone to reproduce exactly the same source build and HammerDB test themselves.

A complete VillageSQL provider therefore needs:

```text
vsqlci.tcl
clone/build/install/start/test support
cifix support
CI validation
VillageSQL provider directories
normal database configuration
I/O-intensive database configuration
```

with configuration files equivalent to the existing MySQL files, for example:

```text
vsql.cnf
vsqlio.cnf
```

This is a substantial part of the HammerDB open source database integration, not simply a future convenience feature.

If the CI provider is definitely intended for a follow-up PR, VillageSQL should not yet be exposed in Pipelines. The Pipelines exposure should be added when the provider exists.

### 4. Preserve the TPROC-C New Order invalid-item rollback semantics

The optimised VillageSQL `NEWORD` implementation batches the order-line data into a JSON array and expands it through `JSON_TABLE`.

That set-based approach is useful and is conceptually similar to the PostgreSQL array/`UNNEST` approach.

However, the deliberate invalid-item case has changed.

HammerDB deliberately generates item:

```text
100001
```

for the rollback case.

In the existing implementation the item lookup finds no row, the `NOT FOUND` handling executes and the entire New Order transaction rolls back.

The VillageSQL implementation instead uses `LEFT JOIN` against `item` and `stock`. Its comments describe retaining the invalid order line and allowing the transaction to continue and commit.

That changes the transaction semantics.

Passing the four HammerDB consistency checks does not prove that this specific New Order rollback requirement has been preserved.

The new set-based implementation should retain the deliberate invalid-item rollback behaviour.

### 5. Use the normative stock quantity rule while refactoring `NEWORD`

The new VillageSQL implementation is already substantially refactoring New Order into a set-based JSON/`JSON_TABLE` implementation, so this is a good opportunity to follow the normative TPC-C transaction definition rather than carry forward the long-standing discrepancy from the Appendix A example.

TPC-C Clause 2.4.2.2 specifies:

```text
S_QUANTITY >= OL_QUANTITY + 10
```

whereas the Appendix A example uses:

```text
S_QUANTITY > OL_QUANTITY
```

HammerDB historically followed the Appendix A form in several implementations. We have previously avoided changing this opportunistically because even a small expression change affected the SQL Server execution plan.

We do have an alternative SQL Server implementation that uses the corrected stock quantity rule with a suitable execution plan, but we have not switched the default implementation because of backward compatibility.

However, VillageSQL is already replacing the row-by-row New Order logic with a new set-based JSON-array implementation, so there is no existing VillageSQL execution plan to preserve here.

As we are refactoring this logic anyway, I think the VillageSQL implementation should use the normative rule:

```sql
CASE
    WHEN s.s_quantity >= ol.q + 10
    THEN s.s_quantity - ol.q
    ELSE s.s_quantity - ol.q + 91
END
```

This should be treated as a deliberate correctness improvement as part of the refactoring rather than reproducing the Appendix A example behaviour.

### 6. The default SSL settings do not work with the official VillageSQL Docker image

The VillageSQL configuration in #920 defaults to:

```text
vsql_ssl false
```

We tested against the current official:

```text
villagesql/server:latest
```

which identified itself as:

```text
8.4.11-villagesql-0.0.6
```

The server was started with:

```bash
docker rm -f vsql-test 2>/dev/null || true

docker run -d \
  --name vsql-test \
  -e MYSQL_ROOT_PASSWORD=mysql \
  -p 3307:3306 \
  villagesql/server:latest
```

and verified with:

```bash
docker exec vsql-test \
  mysql -uroot -pmysql \
  -e "SELECT VERSION(), @@version_comment;"
```

which returned:

```text
VERSION()                     @@version_comment
8.4.11-villagesql-0.0.6       Source distribution
```

Attempting the HammerDB schema build over TCP with the default non-SSL configuration failed with:

```text
Vuser 1:the tcp connection to 127.0.0.1:3307 could not be established
Error in Virtual User 1: mysqlconnect/db server:
Authentication plugin 'caching_sha2_password' reported error:
Authentication requires secure connection.
```

To connect successfully using the current official Docker image we had to copy its CA certificate:

```bash
sudo mkdir -p /opt/vsql-certs

sudo docker cp vsql-test:/var/lib/mysql/ca.pem /opt/vsql-certs/ca.pem

sudo chmod 755 /opt/vsql-certs
sudo chmod 644 /opt/vsql-certs/ca.pem

ls -l /opt/vsql-certs
```

Then restart `hammerdbcli` and configure SSL:

```tcl
dbset db vsql
dbset bm TPROC-C

diset connection vsql_host 127.0.0.1
diset connection vsql_port 3307
diset connection vsql_socket null

diset connection vsql_ssl true
diset connection vsql_ssl_linux_capath /opt/vsql-certs
diset connection vsql_ssl_ca ca.pem
diset connection vsql_ssl_two_way false

diset tpcc vsql_user root
diset tpcc vsql_pass mysql
diset tpcc vsql_dbase tpcc

diset tpcc vsql_count_ware 10
diset tpcc vsql_num_vu 5

buildschema
```

A user following the official VillageSQL Docker setup and the default HammerDB VillageSQL settings will therefore hit an authentication failure over TCP.

This is particularly important for Windows because the VillageSQL implementation disables Unix socket use on Windows, so Windows clients necessarily connect over TCP.

At minimum, the required SSL setup should be clearly documented. Preferably, the default HammerDB VillageSQL configuration should work directly with the current official VillageSQL distribution.

### 7. The GUI transaction counter uses the MySQL colour instead of the VillageSQL colour

When running VillageSQL, the live GUI transaction counter uses the **MySQL orange colour** rather than the assigned **VillageSQL purple colour**.

**Screenshot 2: GUI transaction counter, incorrect MySQL colour**

Image

The web service already handles the VillageSQL colour correctly, as shown by the same database being rendered using the assigned **VillageSQL purple**.

**Screenshot 3: Web service, correct VillageSQL colour**

Image

The issue therefore appears to be limited to the **GUI transaction-counter colour selection**. VillageSQL should use its own configured colour consistently in the GUI, as it already does in the web service.


## Summary

The direction of #920 is good. The build works, VillageSQL is established as its own HammerDB database, the `mysqltcl` foundation can be reused, TPROC-C builds and runs successfully, and the consistency check passes.

Before approval, I think the following seven points should be addressed:

1. Correct VillageSQL Metrics so that it uses the VillageSQL configuration.
2. Remove the inherited OceanBase functionality.
3. Implement the full VillageSQL CI provider, or remove the premature Pipelines exposure until the provider exists. I strongly recommend that this functionality be included as it provides the ability to fully automate the benchmark process.
4. Preserve the intentional New Order invalid-item rollback.
5. Use the normative `S_QUANTITY >= OL_QUANTITY + 10` stock rule while refactoring `NEWORD`.
6. Resolve or clearly document the SSL requirements for the official VillageSQL Docker distribution.
7. Fix the GUI transaction counter so VillageSQL uses the VillageSQL colour rather than the MySQL colour.

I have not tested Active Session History yet and will do so when getting closer to full implementation.

Also to note there is another Pull Request which will be submitted soon after testing to fix an issue where a single pipeline run after a profile pipeline inherits the profiles profile_id (a single profile_id should be 0). This is a simple fix but should be applied to VillageSQL as well. Once submitted the PR should be obvious or please ask.

sm-shaw:
Also to add for the CI changes vsql.cnf and vsqlio.cnf would be fetched from the HammerDB website and is not committed into the code so would need to be provided separately. Examples are already there for MySQL.

sm-shaw:
This is the additional PR that fixes a profile id inheritance bug that should be included for VillageSQL https://github.com/TPC-Council/HammerDB/pull/925

tomas-villagesql:
Thanks Steve for the in-depth review.  I'm working on addressing the issues.  One question though.  My analysis on the PGSQL version that uses UNNEST, does not roll back (afaicu), so that is why I chose to do the same in the JSON_TABLE version.

tomas-villagesql:
On **review point 6 (default SSL / caching_sha2_password over TCP)** — I could not reproduce this as a VillageSQL-specific problem, and the behaviour I did see points at `caching_sha2_password`'s standard first-auth handling rather than anything specific to the VillageSQL server or the `vsql_ssl false` default. I think it needs a little more joint digging to pin down, so sharing what I found.

**Server configuration is identical to stock MySQL.** Official `villagesql/server:latest` (8.4.11-villagesql-0.0.6) and stock `mysql:8.4` report the same relevant settings:

```
require_secure_transport                     = 0
caching_sha2_password_auto_generate_rsa_keys = 1
Rsa_public_key                               = <present>
```

**The failure reproduces against stock MySQL too — and it is cache/timing dependent.** `caching_sha2_password` requires a secure channel (TLS *or* the RSA public-key exchange) for a user's *first* authentication, then caches the credential so subsequent non-SSL connections succeed. Against a freshly started stock `mysql:8.4`, before the credential cache is warm, a non-SSL TCP connection gives exactly the error in your report:

```
Authentication plugin 'caching_sha2_password' reported error:
Authentication requires secure connection.
```

Once root has authenticated once (e.g. the container's own init/socket connection, or any prior TLS/RSA connection), the same non-SSL TCP connection then succeeds — against both stock MySQL and `villagesql/server:latest`. In my environment mysqltcl (`-ssl false`) connects non-SSL over TCP to both servers.

Because the same error appears against stock MySQL, this does not look VillageSQL-specific, and the `vsql_ssl false` default matches `mysql_ssl false` / `maria_ssl false`.

Two things would help locate the real cause:

1. Which `libmysqlclient` version is bundled in the HammerDB build you tested? (The non-TLS RSA-public-key path for `caching_sha2_password` varies between client-library versions — some require an explicit `get-server-public-key` request before they will use the key. My mysqltcl here links `libmysqlclient.so.21`; the HammerDB distribution bundles `.so.24`.)
2. Does the **same** non-SSL TCP connection fail for you against a freshly started **stock `mysql:8.4`** container?

Question 2 is the key fork:
- If stock MySQL fails the same way, this is a general HammerDB/mysqltcl `caching_sha2_password` first-auth matter (client-side RSA-key handling), independent of VillageSQL, and any fix would apply to the MySQL engine too.
- If stock MySQL works but `villagesql/server:latest` does not — despite identical server settings — then the difference is in our VillageSQL Docker image, and that is where I would investigate next (how the image initialises auth / warms the credential cache, and how the client negotiates against it).

tomas-villagesql:
Thanks again Steve, I now pushed a number of commits to the PR:

-  Metrics use VillageSQL config
- Remove OceanBase (2 commits)
- NEWORD invalid-item rollback
- Normative stock rule (also changed the related place uncommented by you)
- GUI counter color

Not addressed yet:
- the ssl issue, needs further investigation on the root cause and how behavior differs vs stock mysql Docker
- CI being worked upon in a separate patch

sm-shaw:
Thanks Tomas, you are correct. I checked the current PostgreSQL `UNNEST` implementation and it does not roll back the invalid-item transaction as intended.

It still deliberately generates item `100001`, but the bulk item lookup uses:

```sql
SELECT array_agg(i_price)
INTO price_array
FROM UNNEST(item_id_array) item_id
LEFT JOIN item ON i_id = item_id;
```

With the invalid item, the `LEFT JOIN` produces a `NULL` element in `price_array` rather than raising `no_data_found`, so the existing rollback handler is not reached.

For context, this behaviour came from the contributed PostgreSQL bulk-processing refactor in #92, merged as `f2127b757177ee882c1815e6205ddc2278d647f5`. #115 subsequently built on that `UNNEST` implementation.

The PostgreSQL change should be small. For example, after creating `price_array` we can detect the failed lookup and roll back:

```sql
IF array_position(price_array, NULL) IS NOT NULL THEN
    ROLLBACK;
    RETURN;
END IF;
```

I also checked Oracle, and that implementation does preserve the intended approximately 1% New Order rollback:

```sql
IF DBMS_RANDOM.value < 0.01 THEN
    dbms_output.put_line('Rolling back');
    ROLLBACK;
ELSE
    COMMIT;
END IF;
```

So we will keep the rollback behaviour in the new VillageSQL implementation and I add the corresponding PostgreSQL correction targetted in the same HammerDB release.

For SSL this is not a blocker for the PR and is almost certainly the same in MySQL. As discussed at Percona Livve, I think this is a good example of where it may make sense for me to update the `mysqltcl` driver rather than require VillageSQL users to work around the client behaviour.

At the moment HammerDB treats SSL as a simple enabled/disabled option and, when enabled, expects the CA configuration to be supplied explicitly. For the current official VillageSQL Docker image that means a new user has to extract `ca.pem` from the container and configure the CA path before `caching_sha2_password` will accept the TCP connection.

A better approach would be for the driver to support automatic/preferred SSL negotiation, so HammerDB can establish a secure connection when the server supports it without requiring the user to manually manage the Docker CA certificate.

I will look at this on the HammerDB/driver side, extending the driver with a more flexible SSL mode would improve the experience for VillageSQL, MySQL and MariaDB.

When the CI patch is included, I will re-review. 

Thanks again!

tomas-villagesql:
The CI patch is now up as #930 (successor to the closed #921, which couldn't be reopened after its branch was rebased onto the review fixes here).

#930 adds the VillageSQL CI provider — the `<VillageSQL>` block in `config/ci.xml`, the `villagesql_*` pipeline step procs in `src/villagesql/vsqlci.tcl` (aliased to the `$rdbms`-driven `mysql_*` implementations), the label/whitelist wiring in `src/generic/genci.tcl`, and the #925 profile-id fix applied to the VillageSQL profile runner.

It's stacked on this PR (contains these engine commits plus the CI-specific ones), so #920 first, then #930. Verified end-to-end on Linux from a source build: `cipush` runs clone → build → package → install → init → start against the VillageSQL source and brings up a working server.

Thanks again for the thorough review.

sm-shaw:
Thanks again for the updates. I have now retested #920 against the current official VillageSQL container and the engine integration is very close.

The current review status is:

### 1. VillageSQL Metrics / ASH configuration

**Not fixed yet and still a blocker.**

The Metrics Options dialog correctly shows the VillageSQL connection settings, including:

```text
Host: 127.0.0.1
Port: 3307
CAPath: /opt/vsql-certs
CA: ca.pem
User: root
```
e.g.
image


However, starting Database Metrics still enters the MySQL connection path and attempts:

```text
Waiting for Connection to MySQL for Database Metrics...
Metrics Connecting to host:127.0.0.1 port:3306
Database Metrics Error: MySQL connection to 127.0.0.1:3306 failed
```

So the VillageSQL dialog is using the correct configuration, but the Metrics backend is still resolving through the MySQL connection settings.

This is the one remaining functional blocker I have reproduced in #920.

### 2. Remove inherited OceanBase functionality

**Fixed.**

The current VillageSQL implementation no longer carries the OceanBase-specific transaction-counter configuration, and the VillageSQL transaction counter now starts and runs correctly.

### 3. VillageSQL CI provider / Pipelines

**Moved to #930.**

The CI provider is now being implemented in the separate stacked PR #930, so I am treating the CI implementation itself separately from the engine review here.

#930 currently has its own review items around package selection and release tag naming.

### 4. Preserve the TPROC-C invalid-item rollback

**Fixed.**

The current `NEWORD` implementation explicitly detects the deliberate invalid item `100001` and forces the transaction rollback path.

### 5. Use the normative stock quantity rule

**Fixed.**

The current implementation now uses the normative rule:

```sql
s_quantity >= ol_quantity + 10
```

### 6. SSL / `caching_sha2_password`

**Resolved as non-blocking for #920.**

VillageSQL works correctly over TLS against the official container.

The non-TLS first-authentication behaviour appears to be general MySQL/client `caching_sha2_password` behaviour rather than something specific to VillageSQL. So we will provide a fix for this.

### 7. GUI transaction counter colour

**Fixed and verified.**


Screenshot 2026-09-15 174830
Screenshot 2026-09-15 175114

The VillageSQL transaction counter now uses the assigned VillageSQL purple colour in the GUI.

### Additional testing completed

I have also now verified:

- TPROC-C schema build
- TPROC-C workload execution
- VillageSQL transaction counter
- CPU and I/O metrics under load
- TLS connectivity against VillageSQL 8.4.11 / 0.0.6
- TPROC-H SF1 build and schema check

So #920 is now down to one remaining issue:

**VillageSQL Database Metrics / ASH still uses the MySQL connection path rather than the VillageSQL connection configuration.**

Once that is fixed and retested, I think #920 will be ready for approval.

tomas-villagesql:
Thanks Steve for the quick turnaround. I now believe I addressed the remaining 1. VillageSQL Metrics / ASH configuration.  And I followed the pattern you suggested for the ci patch and updated the common file to handle config name variations. (I don't have a working gui locally so it has only been unit tested)

sm-shaw:
I have now tested both the automated/CLI paths and the GUI paths.

For the automated CI testing:

- `Single` completed successfully.
- `Compare` completed successfully using VillageSQL `0.0.3` and `0.0.4`.

vsql1

vsql2

That confirms the current Compare implementation works with those tags. I would not assume that guarantees compatibility with `release/0.0.6` or future tag naming, so this may need a small follow-up once the VillageSQL release naming convention has settled.

For the GUI testing, the workload and transaction counter worked, but Database Metrics/ASH initially failed with:

```text
Unmatched Background Error - invalid command name "find_config"
    while executing
"find_config $rdbms "
    (procedure "connect_to_mysql" line 11)
    invoked from within
"connect_to_mysql"
    (procedure "::mysqlmet::mysqlmetrics" line 27)
    invoked from within
"vsqlmetrics"
    (procedure "metrics" line 64)
    invoked from within
"metrics"
    invoked from within
".ed_mainFrame.buttons.dashboard invoke "
    invoked from within
"if { ![ string match [ .ed_mainFrame.treeframe.treeview state ] "disabled focus hover" ] } { .ed_mainFrame.buttons.dashboard invoke } "
    (command bound to event)
(HammerDB-6.0) %
```

The provider-aware Metrics change in #920 uses `find_config` and `find_prefix`, but those helpers were not available in the GUI execution path.

I have tested the fix successfully. Database Metrics and ASH now work correctly with VillageSQL in the GUI.

I do not have write permission to `villagesql/HammerDB`, so I have put the clean tested fix here:

https://github.com/HammerDBLtd/HammerDB/commit/8176901

Please cherry-pick that commit onto `villagesql-engine` for #920 to fix the above error.

Once that cherry-pick is done, the remaining sequence is:

1.#920 is good to merge, I will approve it.
2. Andy can then merge #920.
3. Once #920 has been merged, please rebase `villagesql-ci` / #930 onto the updated `master`. I cannot do this rebase myself for the same reason as the cherry-pick because I do not have write access to the `villagesql/HammerDB` fork.
4. I can then do the final #930 check and approve #930.
5. Andy can then merge #930.
6. After #920 and #930 are complete, I would handle the SSL/no-CA work separately for VillageSQL, MySQL and MariaDB to ensure that SSL = true doesn't explicitly require the files to be present.

tomas-villagesql:
thanks, cherry pick is done.  I did not re-test after the change

sm-shaw (approved):
Retested successfully across the GUI and automated/CLI paths, including Metrics/ASH with the final fix. 
I already thoroughly tested this fix - screenshot in #930
Approved.

abondvt89 (approved):

abondvt89:

sm-shaw:
PR #937 fixes 6. The default SSL settings do not work with the official VillageSQL Docker image

PR #924, opened 2026-09-13
Fix transaction counter graph clipping

sm-shaw:
## Description

Fix a small rendering issue in the GUI transaction counter where the 5-pixel graph stroke could extend slightly beyond the right-hand plot boundary when the first transaction sample was drawn at a steep angle.

The fix reuses the existing `pad_stroke` value and moves only the final x-coordinate of the visible polyline inside the plot area.

The left-hand edge, graph scaling, autorange, axes and gradient fill are unchanged.

## Testing

Tested in the GUI with the transaction counter starting from zero and with subsequent samples updating across the graph.


abondvt89 (approved):

abondvt89:

PR #925, opened 2026-09-13
Fix single pipeline runs inheriting a previous profile ID

sm-shaw:
## Summary

Fixes Single pipeline runs retaining a previously persisted performance profile ID.

A Single pipeline uses `PROFILEID=0`, but the TPROC-C runner scripts only called:

`jobs profileid $profileid`

when `PROFILEID > 1`.

Because the jobs profile ID is persisted, a Single run executed after a Profile run could inherit the previous non-zero profile ID and be incorrectly associated with that profile.

## Change

Always apply the validated `PROFILEID` before the benchmark run.

- Single run: `jobs profileid 0`
- Profile run: existing non-zero profile ID behaviour is unchanged

The change is applied consistently to the Tcl and Python TPROC-C runners for:

- PostgreSQL
- MySQL
- MariaDB

## Testing

Tested using the web service with PostgreSQL.

1. Ran a Profile pipeline with `profile_id = 1000`
2. Ran a Single pipeline immediately afterwards
3. Confirmed the Single pipeline completed with:

    `pipeline: single_c`

    `profile_id: 0`

    `status: COMPLETE`

4. Confirmed the Single benchmark job was not included in Profile 1000

This verifies that a Single pipeline no longer inherits the profile ID from a previous Profile run.


abondvt89 (approved):

abondvt89:

PR #929, opened 2026-09-14
Propagate MySQL schema loader errors to the build monitor

### LINKED ISSUE #928: MySQL schema build monitor can wait indefinitely after a loader error
**Guidance**

**Describe the bug**
In MySQL TPROC-C and TPROC-H multi-VU schema builds, a worker exception can leave the monitor waiting indefinitely for every worker to report `done`.

The generated `do_tpcc` / `do_tpch` procedures let exceptions propagate without consistently setting `application abort`. Separately, the monitor's `while 1` loop checks worker completion but does not check `application abort`. Thus even a failure path that sets abort does not stop this loop. A monitor setup exception can also leave workers waiting for READY instead of notifying them immediately.

Affected source: `src/mysql/mysqloltp.tcl` and `src/mysql/mysqlolap.tcl` on master `f5fd6f0c164ad7c4f426d097ac87a8ffa0cadd49`.

**To Reproduce**

1. Use the original master MySQL build generators with more than one loader VU (for TPROC-C, also configure at least two warehouses).
2. Inject a connection exception in a worker after the READY state is available. Observe that the original error propagates but `application abort` remains 0 and the worker never reports `done`.
3. Exercise the monitor with a failed worker and `application abort` set to 1. The original monitor still keeps polling instead of exiting.
4. Run the deterministic regression in [tests/mysql-loader-errors.tcl](https://github.com/oceanbase/HammerDB/blob/codex/mysql-loader-abort/tests/mysql-loader-errors.tcl) against original and corrected source. It extracts the actual generated loader procedures and uses simulated database calls, shared state and bounded polling, so no database or indefinite wait is required.

```text
tclsh tests/mysql-loader-errors.tcl
```

Tcl 8.6+ is required; the script can also run using `hammerdbcli auto`. `HAMMERDB_SOURCE_ROOT` selects an alternate source checkout for baseline comparison.

**Expected behavior**
An exception in either a worker or the monitor should set the shared abort flag while preserving the original error message and error code. The monitor should observe abort and exit on its next poll, instead of waiting for a failed worker to report done. Successful builds should retain their existing completion behavior.

**Screenshots**
Not applicable; this is a CLI regression with textual test results.

 - Version: 6.0
 - Build: Release download CLI on AWS; original and corrected generated loader procedure bodies were loaded into isolated Tcl interpreters for comparison.

 - UI: CLI

 - Server OS: Not applicable; no live database server was used.
 - Client OS: Ubuntu 24.04, Linux x86-64 on AWS EC2

**Database Server (please complete the following information):**
 - Database name: MySQL workload, using deterministic database-result fixtures
 - Database Release Version: Not applicable; no live database server was used.

**Database Client (please complete the following information):**
 - Database client name: Mocked MySQL Tcl commands; no live database connection was used.

**Additional context**
AWS validation with the 6.0 CLI: 10 tests, baseline 4 passed / 6 failed; corrected procedures 10 passed / 0 failed. For each workload the suite covers worker connection errors, monitor setup errors, monitor abort handling, successful workers and successful monitors. The polling fixture terminates after a bounded number of iterations so a regression cannot hang the test runner.

The change directly handles errors inside the existing generated MySQL loaders. It introduces no OceanBase dependency and no shared runtime module. It does not change the READY timeout or add reconnect/retry behavior.

Related reports #734 and #309 concern lost connections after data loading, and #857 concerns a READY timeout. This report addresses propagation of loader exceptions and the monitor ignoring abort. Windows, GUI execution and a live MySQL server regression have not been validated.

jackysp:
Fixes #928.

The generated MySQL TPROC-C and TPROC-H schema loaders can fail without setting the shared `application abort` flag, leaving the monitor waiting for a worker that will never report `done`. The monitor also ignores this flag while polling. This change wraps both generated loader bodies with error handling that sets `application abort`, preserves the original error and error code, and checks the flag in each monitor loop.

This is a standalone MySQL fix based on master. It has no OceanBase dependency, adds no runtime module, and leaves READY timeouts and reconnect behavior unchanged.

Validation used 10 deterministic regression tests on AWS Linux x86-64 with the HammerDB 6.0 CLI. The original generated loader bodies passed 4/10 tests and failed 6/10; the corrected bodies passed 10/10. Tests cover worker and monitor connection failures, preserved error messages and codes, bounded monitor polling after abort, and normal worker/monitor completion for both workloads. Database calls and shared state are simulated; Windows, GUI, and live MySQL server execution are not covered.

Run `tclsh tests/mysql-loader-errors.tcl` with Tcl 8.6 or later, or use `hammerdbcli auto tests/mysql-loader-errors.tcl`. The test reads the source checkout and extracts the generated loader procedures; `HAMMERDB_SOURCE_ROOT` can select a baseline checkout.


jackysp:
Fixes #928.

A MySQL schema loader can fail without setting the shared abort flag, leaving the monitor waiting for a worker that will never report done. The monitor also ignores abort while polling. Wrap the generated TPROC-C and TPROC-H loader bodies with error handling that sets abort and rethrows the original error, and check abort in each monitor loop.

This is a standalone MySQL fix based on master. It has no OceanBase dependency, adds no runtime module, and leaves READY timeouts and reconnect behavior unchanged.

Validation: ran 10 deterministic regression tests on AWS Linux x86-64 using the HammerDB 6.0 CLI. Original generated loader bodies: 4 passed / 6 failed. Corrected bodies: 10 passed / 0 failed. Tests exercise worker and monitor connection failures, preserved error messages/codes, bounded monitor polling after abort, and normal worker/monitor completion for both workloads. Database calls and shared state are simulated; Windows, GUI and live MySQL server runs are not covered.

Run `tclsh tests/mysql-loader-errors.tcl` with Tcl 8.6+, or use `hammerdbcli auto tests/mysql-loader-errors.tcl`. The test reads the source checkout and extracts the generated loader procedures; `HAMMERDB_SOURCE_ROOT` can select a baseline checkout.

sm-shaw (changes requested):
So this is a long known issue/feature, however in my experience loader thread errors are rare and the inconvenience of checking the state has not taken priority over simplicity. However with larger and larger builds it is recognised that this can be revisited as then be defintion failures could be more likely.  

If so, looking at the implementation across HammerDB, I think this should be treated as a **generic schema-loader state-handling improvement**, not a MySQL-only fix.

The multi-threaded schema loaders already maintain per-worker state in `common thrdlst` and the monitor watches those states:

```tcl
idle
active
done
```

This same pattern is used across the TPROC-C / TPROC-H loaders for MySQL, MariaDB, VillageSQL, PostgreSQL, Oracle, SQL Server and Db2.

That gives us a cleaner way to address the issue than overloading the global `application abort` flag.

I would suggest extending the existing worker state model to include a failure state, for example:

```tcl
idle
active
done
error
```

A loader worker that encounters an exception should set its own state to `error` while preserving and reporting the original exception. The monitor already examines every worker state, so it can detect an `error` state immediately rather than waiting indefinitely for every worker to reach `done`.

Conceptually:

```tcl
try {
    # worker load operations

    tsv::lreplace common thrdlst $myposition $myposition done
} on error {message options} {
    tsv::lreplace common thrdlst $myposition $myposition error
    return -options $options $message
}
```

and the existing monitor loop can be extended along the lines of:

```tcl
set errcnt 0

for {set th 2} {$th <= $totalvirtualusers} {incr th} {
    switch [tsv::lindex common thrdlst $th] {
        idle   { incr idlcnt }
        active { incr lvcnt }
        done   { incr dncnt }
        error  { incr errcnt }
    }
}
```

The monitor can then terminate the build appropriately when one or more workers report `error`.

This keeps two different concepts separate:

```text
application abort = application/user cancellation

worker error       = schema loader worker failed
```

The current PR instead makes `application abort` mean both cancellation and loader failure:

```tcl
if {[tsv::get application abort]} {
    error "Schema build aborted after a loader failure"
}
```

That is problematic because `application abort` is already used throughout HammerDB as a normal stop/cancellation signal.  i.e. pressing the stop button. A user-requested stop therefore cannot safely be interpreted as a loader failure.

The outer `try` around the complete `do_tpcc` / `do_tpch` procedure also changes the error handling semantics of considerably more than the loader workers themselves. I would prefer the change to be scoped to the particular worker/monitor state transitions that need improving.

There is a similar consideration for the monitor itself. If the monitor fails before setting the existing `application load` state to `READY`, that should be represented explicitly so waiting workers can exit rather than waiting for the existing timeout. For example, the existing `WAIT` / `READY` state could potentially have an `ERROR` state rather than using `application abort` to infer what happened.

Although #928 was raised against MySQL, this is not really a MySQL-specific design issue. The same `idle` / `active` / `done` monitoring model is used by the other HammerDB multi-threaded schema loaders. If we improve this mechanism, I would rather do it consistently across the supported databases than add special error semantics to MySQL alone.

This has historically been more of an inconvenience than a serious functional problem because schema-loader failures are relatively uncommon, which is why it has not previously been prioritised. However, if we are going to address it now, I think it is worth doing through the existing per-worker state mechanism rather than introducing a second meaning for the global abort flag.

### Changes required

- Use the existing per-worker `common thrdlst` state mechanism to represent a loader failure, for example with an `error` state.
- Preserve the original loader error and error information.
- Extend the monitor to detect the worker `error` state and stop waiting for all workers to report `done`.
- Keep `application abort` as the application/user cancellation mechanism rather than using it to identify loader failures.
- Handle monitor setup failure explicitly so workers waiting for `READY` can exit without waiting for the timeout.
- Apply the solution consistently to the multi-threaded schema loaders rather than making this MySQL-specific.
- Remove the new `tests/mysql-loader-errors.tcl` file. HammerDB does not maintain a top-level Tcl regression-test suite and we intentionally keep the main source tree lightweight. Validation should use the existing build, workload, CI and platform testing processes.

With those changes I think this could be a useful general improvement to the schema-loader state handling as database builds increase in size.

PR #930, opened 2026-09-14
Add VillageSQL CI provider

tomas-villagesql:
> **Stacked on #920** — this branch contains the #920 engine commits plus the CI-specific commits below. Please review/merge #920 first; the CI diff here is `config/ci.xml`, `src/villagesql/vsqlci.tcl`, `src/generic/genci.tcl`, and `scripts/tcl/villagesql/tprocc/vsql_tprocc_run_profile.tcl`. Successor to the closed #921 (that PR could not be reopened after its branch was rebased onto the #920 review fixes).

## Summary

Adds the VillageSQL CI provider — the "separate patch" referenced in #920 and the CI work requested before re-review.

Builds on top of the #920 engine branch (all seven #920 review points are addressed there), then adds the CI-specific pieces:

- **`config/ci.xml`** — a `<VillageSQL>` provider block. VillageSQL is a MySQL 8.4 fork, so it is driven as a raw `mysqld` like the MySQL provider (in-source cmake `RelWithDebInfo`, `WITH_SSL=system`; `bin/mysqld` install/init/start), rather than via its own launcher.
- **`src/villagesql/vsqlci.tcl`** — the pipeline step procs (`villagesql_clone` / `build` / `package` / `install` / `init` / `start` / `ping` / `run_sql` / `start_tests`, plus `profile` / `compare`). These alias to the shared `mysql_*` implementations, which are `$rdbms`-driven (they read `$cidict $rdbms` and `find_prefix $rdbms`), so they operate on the VillageSQL provider with no duplication.
- **`src/generic/genci.tcl`** — the `vsql` → `VillageSQL` label mapping and `vsql.cnf` / `vsqlio.cnf` config whitelisting.
- **`scripts/tcl/villagesql/tprocc/vsql_tprocc_run_profile.tcl`** — includes the profile-id inheritance fix from #925 (always apply the validated PROFILEID; a Single run sets `jobs profileid 0` rather than inheriting a previous Profile run's id).

## Testing

Verified end-to-end on Linux against a from-source HammerDB build. `cifix` sets up the provider tree and processes VillageSQL as a first-class provider; `civalidate` passes; `cipush <ref> single C vsql` runs the pipeline through **clone → build → package → install → init → start**, producing a working server built from the VillageSQL source (`SELECT @@version` served over the configured socket).

Notes:
- `vsql.cnf` / `vsqlio.cnf` are host-supplied (not committed), the same convention as the MySQL provider's `my.cnf` / `myio.cnf`.
- The ref is chosen at trigger time exactly as for the other engines (e.g. `cipush refs/heads/main ...`); nothing VillageSQL-specific about ref handling.


sm-shaw (commented):
Thanks for the update. This is very close now, and the overall approach of reusing the MySQL CI implementation for VillageSQL makes sense.

I have now tested the VillageSQL `Single` pipeline against the `0.0.5` tag. The build and package steps complete successfully, but there are still a couple of MySQL-specific naming assumptions in the shared CI code that prevent VillageSQL from completing the pipeline.

The first issue is package selection during install.

VillageSQL produces more than one `.tar.gz` artifact, for example:

```text
mysql-8.4.10-villagesql-0.0.5-dev-a2a825d0fe4-linux-x86_64.tar.gz
villagesql-extension-sdk-mysql-8.4_0.0.5-dev.tar.gz
```

The shared `mysql_install` implementation currently does:

```tcl
set files [glob -nocomplain -directory $local_dir *.tar.gz]
set first_file [file tail [lindex $files 0]]
```

so it simply installs the first tarball returned.

In my test this selected the VillageSQL extension SDK rather than the server package. The extraction itself succeeds, so the CI reports:

```text
INSTALL SUCCEEDED
```

but the following init step then fails because there is no installed MySQL/VillageSQL server tree:

```text
DB init failed: no mysql-* directories under /opt/hammerdb-ci/providers/villagesql/installs/ci_1_0.0.5
INIT FAILED
```

The install step therefore needs deterministic package selection. For VillageSQL it should select the actual server package rather than the first `.tar.gz` artifact.

However after install this should then match because it switches to a mysql prefix in the extracted file which matches the current mysql implementation. 
```
 tar -tvf mysql-8.4.10-villagesql-0.0.5-dev-a2a825d0fe4-linux-x86_64.tar.gz | more
-rw-r--r-- ubuntu/ubuntu   679 2026-09-15 16:16 mysql-8.4.10-villagesql-0.0.5-dev-a2a825d0fe4-linux-x86_64/READ
ME.router
```


There is a similar naming incompatibility in the Compare path.

`villagesql_compare` currently reuses `mysql_compare`, but `mysql_compare` determines the previous release using:

```tcl
git -C $repo tag -l "mysql-*" --sort=v:refname
```

This does not match the current VillageSQL release tags, for example:

```text
0.0.5
release/0.0.6
```

so a normal VillageSQL tag comparison will fail even though the rest of the Compare logic is reusable. (If a commit is chosen it should work). 

I think the underlying issue is the same in both cases: most of the shared MySQL CI implementation is provider-driven, but a few naming conventions are still hard-coded to MySQL.

Rather than duplicating the MySQL implementation, I would suggest making these naming assumptions configurable or provider-aware, particularly:

- server package selection
- installed server directory matching
- release tag matching for Compare

On the VillageSQL side, the release naming is also a little inconsistent at present. `0.0.5` is quite generic, while `release/0.0.6` introduces a different convention. A consistent VillageSQL-specific naming scheme such as:

```text
villagesql-0.0.5
villagesql-0.0.6
```

or similar would make automated CI handling clearer and avoid ambiguity with inherited MySQL tags.

Similarly with the package install:

`villagesql-0.0.5-dev-a2a825d0fe4-linux-x86_64.tar.gz`

would be more interoperable with 3rd party tools than the current:

`mysql-8.4.10-villagesql-0.0.5-dev-a2a825d0fe4-linux-x86_64.tar.gz`

The good news is that the core integration appears sound. `cifix` completes successfully, the VillageSQL configuration is recognised correctly, the source builds, and the expected server package is produced. The remaining problems look to be compatibility issues around artifact and tag naming rather than a problem with the overall CI design.

I would therefore treat these as blockers for #930, they look quite contained to fix, however naming consistency between a mysql prefix, villagesql prefix and no prefix would really help identify the correct patterns to match for automation.

sm-shaw (changes requested):
Requesting changes because the remaining blockers come from inconsistent naming across the VillageSQL build and release artifacts: some use villagesql-*, some retain mysql-*, and some release tags have no product prefix at all. That inconsistency makes it hard for the shared CI logic to identify the correct server package and the correct release tags reliably. The overall integration looks sound, but the naming needs to be made consistent or the matching logic made explicitly provider-aware before approval.

tomas-villagesql:
Thanks Steve, I've fixed the blockers and tested end-to-end — clone through benchmark, all green — against the VillageSQL release tags release/0.0.6, 0.0.1, 0.0.2, and I re-ran stock MySQL to confirm the shared-code changes leave the existing provider unchanged. The three shared-CI naming assumptions are now provider-aware with fallbacks, so MySQL/MariaDB/PostgreSQL are unaffected:

- Install now reads install/package_pattern — VillageSQL's build emits two tarballs (server + extension SDK), and the pattern correctly picks the server and skips the SDK (this was the INSTALL SUCCEEDED → INIT FAILED case).
- Compare now reads install/tag_pattern instead of the hardcoded mysql-*.
- Clone uses the ref name verbatim, so multi-segment tags like release/0.0.6 clone correctly (the old file tail reduced it to 0.0.6 and failed).

You were right that testing against a dev branch hid this — a single-tarball branch never exercised the multi-artifact or tag paths. Agree too that the VillageSQL-side naming (0.0.5 vs release/0.0.6, the mysql-…-villagesql-… package name) should be made consistent; the matching here is provider-aware so it isn't blocked on that.

tomas-villagesql:
I also changed so that villagesql uses -O3 in the build like stock mysql, but notable I'm not seeing any difference in performance in my runs

tomas-villagesql:
Also, this PR needs a rebase as there was a change in 920 which this PR builds on. Not sure if GitHub will figure it out automatically or not when 920 goes in.

sm-shaw:
Metrics/Active Session History now working for VillageSQL after additional commit. Comments in #920 to cherry-pick this commit. #920 will then be ready to approve. After #920 merge please rebase #930 onto master (we don't have permissions), then we will merge #930 and be good to go. 

vsql4
vsql5

tomas-villagesql:
thanks, will tidy up this one once 920 is in

sm-shaw:
#920 is now approved and ready to merge.

I have checked the current branch relationship and `villagesql-ci` is exactly 6 commits ahead and 0 behind the final #920 `villagesql-engine` head, so it already contains the final #920 changes.

Once #920 is merged, I will re-check #930 against the updated `master`.

If it resolves cleanly to just those six CI commits, no rebase should be necessary and I will complete the final review and approve #930 for merge.

If it does not resolve cleanly against the updated `master`, please rebase `villagesql-ci` onto `master` and I will review and approve after that.

sm-shaw (approved):
Final check complete after #920 merge.
#930 now reduces cleanly to the expected 6 CI commits / 5 files against master no rebase required. Approved and ready to merge.

abondvt89 (approved):

abondvt89:

PR #936, opened 2026-09-18
Fix PostgreSQL New Order invalid-item rollback

sm-shaw:
The PostgreSQL TPROC-C New Order implementation using arrays and `UNNEST` can lose the required invalid-item rollback path because the `LEFT JOIN` item lookup returns a `NULL` price for item `100001` rather than raising `NO_DATA_FOUND`.

This change restores the TPROC-C rollback behaviour by checking the generated `price_array` for `NULL` and raising `NO_DATA_FOUND` when an invalid item is present. The existing exception handler then rolls back the complete New Order transaction.

The check is applied to both PostgreSQL New Order variants.

No measured impact on performance with the fix but added for completeness.

The added check is:

```sql
IF array_position(price_array, NULL) IS NOT NULL
THEN
    RAISE NO_DATA_FOUND;
END IF;


abondvt89 (approved):

abondvt89:

PR #937, opened 2026-09-21
Allow MySQL-family TLS without explicit CA files

sm-shaw:
Allow MySQL, MariaDB and VillageSQL TLS without explicit CA files

## Background
PR #920 added VillageSQL as a first-class HammerDB database engine for TPROC-C and TPROC-H. VillageSQL is MySQL 8.4 based, so the integration uses the existing `mysqltcl` client library and follows the existing MySQL SSL/TLS handling.

Post-merge testing of #920 against the VillageSQL Docker image exposed a limitation in that handling. When SSL was enabled, HammerDB assumed that a CA path or CA/certificate/key files would also be supplied. However, the current client library supports requesting an encrypted connection with simply:

```tcl
-ssl true
```

without requiring explicit CA files.

As a result, VillageSQL could not use its normal TLS-enabled configuration from HammerDB without providing CA files even though the client and server support TLS without them.

The same SSL option handling is shared conceptually by MySQL and MariaDB, so this PR resolves the issue consistently across all three databases rather than special-casing VillageSQL.

## Changes

When SSL is enabled with one-way TLS and the CA path, CA, certificate and key are all blank, HammerDB now passes:

```tcl
-ssl true
```

to the client library.

This allows the client to establish an encrypted connection without requiring explicit certificate verification files.

The existing SSL-disabled and explicit CA/certificate configuration paths are retained.

SSL options are also rebuilt when the SSL configuration changes within the same HammerDB session. Unchanged settings reuse the existing options, including during `refreshscript`, avoiding repeated SSL validation messages.

Invalid explicit SSL paths no longer leave the `*_ssl_options` variable undefined or silently switch the configuration to a non-SSL connection. The requested SSL options are retained and the client connection reports the SSL configuration failure.

The same handling is applied to TPROC-C and TPROC-H for:

- MySQL
- MariaDB
- VillageSQL

## Testing

Tested with HammerDB 6.1 on Linux.

### VillageSQL

Tested against the VillageSQL Docker image based on MySQL 8.4.

Verified:

- SSL enabled, one-way TLS, no CA path or CA/certificate/key files
- TLS connection established successfully
- TPROC-C schema build completed successfully
- TPROC-C runtime connection completed successfully
- SSL enabled with a valid `ca.pem`
- SSL disabled
- SSL configuration changed between enabled and disabled states in the same HammerDB session
- SSL configuration changed from an invalid CA path to no-CA TLS in the same HammerDB session
- invalid CA path reports the SSL configuration error and the client SSL connection fails without an undefined `vsql_ssl_options` error
- `refreshscript` does not repeat the SSL validation message when the SSL settings are unchanged
- GUI schema builds tested with SSL enabled, SSL disabled and SSL re-enabled with no CA files in the same GUI session

Example successful no-CA TLS connection:

```text
Vuser 1:Ssl_cipher TLS_AES_128_GCM_SHA256
```

### MySQL

Tested against MySQL 8.4 Docker.

Verified one-way TLS with SSL enabled and no CA path or CA/certificate/key files.

```text
Vuser 1:Ssl_cipher TLS_AES_128_GCM_SHA256
```

### MariaDB

Tested against MariaDB 11.4 Docker.

Verified one-way TLS with SSL enabled and no CA path or CA/certificate/key files.

```text
Vuser 1:Ssl_cipher TLS_AES_256_GCM_SHA384
```

This resolves the SSL/TLS compatibility issue exposed by the VillageSQL integration in #920 while applying the fix consistently to the MySQL-family database implementations.


abondvt89 (approved):

abondvt89:

